PR #151, opened 2013-08-18
Fix a problem when your .emacs.d is a symlink.

xiaohanyu:
If you .emacs.d dir is a symlink to another directory, then sp-cheat-sheet
will return nothing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 elpa
+*.elc


rprimus:
Tue Jan 19 01:05:41 GMT 2021

Hi,

I've just tried `M-x sp-cheat-sheet` and got a blank help buffer.  Removing the `file-truename` call produced the `cheat sheet`.

* `~/.emacs.d/` is a symlink, however, `load-history` does not contain/reference its `file-truename`.
* "GNU Emacs 27.1 (build 1, x86_64-apple-darwin19.6.0, Carbon Version 162 AppKit 1894.6)\n of 2020-11-15"

Fuco1:
Could it be that this is an Apple issue?  When I look on load-history on linux, all the file names seem to be expanded.

rprimus:
Sun Mar 28 18:40:56 BST 2021

> ~/.emacs.d/ is a symlink, however, load-history does not contain/reference its file-truename.
>> When I look on load-history on linux, all the file names seem to be expanded.

To clarify:

```
(locate-library "smartparens")
 => "/Users/m/.emacs.d/straight/build/smartparens/smartparens.el"
```
The path above is the path that appears in `load-history`
```
(file-truename (load-library "smartparens"))
=> "/Users/m/dotfiles/.emacs.d/straight/repos/smartparens/smartparens.el"
```

```
: ; ls -lad ~/.emacs.d
lrwx------ 1 m staff 17 Oct 14 17:17 /Users/m/.emacs.d -> dotfiles/.emacs.d
```

rprimus:
Sun Mar 28 18:56:29 BST 2021

Note:

At the time of my original comment,  I was not using `straight`, so there was only 1 symlink: 
`~/.emacs.d -> dotfiles/.emacs.d`.

`straight` introduces an additional symllink:
```
: ; cd .emacs.d
➜  .emacs.d git:(main)
: ; ls -lad straight/build/smartparens/smartparens.el
lrwx------ 1 m staff 59 Mar 28 14:11 straight/build/smartparens/smartparens.el -> /Users/m/.emacs.d/straight/repos/smartparens/smartparens.el
```

Fuco1:
I suppose mapping true-filename over the alist would solve this in the general case.  Although I'm also using straight and the symlinks still seem to be resolved.

I pushed a commit which does this.  Feel free to try it out!

Issue #260, opened 2013-11-30
Method for defining global filters


porterjamesj:
The manual describes how I can add filters to specific pairs, but as far as I can tell there isn't a way to add a *global* filter, which would determine the autopairing behavior of all pairs, before any pair specific filters are run. I'm marginally competent at elisp so if you give me a rough sketch of what would need to be modified I'd be happy to try and implement it :)

Fuco1:
Continues in #261

PR #261, opened 2013-11-30
WIP: add basic implementation of global filters

porterjamesj:
First pass at the feature suggested in #260. Not sure if this is something you want or if there's a better way to do it, just opening up a discussion.


Fuco1:
Since we removed the global config and/or the related settings, I'm going to close this.  Might be the longest opened pull request in github history :blush:

Issue #289, opened 2014-01-24
String-like delimiters and `sp-show-pair-from-inside`


ThibautVerron:
Hello,

Depending on the value of `sp-show-pair-from-inside` I see the following behavior: 

Consider the following text in a latex buffer:

```
$({})$
```

Assume the value of `sp-show-pair-from-inside` is nil (the default).
If the point is before `}` nothing is highlighted.
If the point is before `)` the pair of braces is highlighted.
If the point is before `$` the pair of dollars is highlighted.
If the point is after `$` the pair of dollars is highlighted.

At no point is the pair of brackets highlighted.

If the value of `sp-show-pair-from-inside` is `t`, the pair corresponding to the character after the point is highlighted, as expected.

Is this the intended behavior? Based on the docstring of `sp-show-pair-from-inside`, if set to `nil`, I would expect the pair of brackets to be highlighted when the point is just after them.

(Possibly linked to https://github.com/Fuco1/smartparens/issues/179 )

Thanks!

Fuco1:
Hello folks, Is this issue fixed?

ThibautVerron:
I am not sure that I understand what I wrote in the first post back then. `:)`

Based on my observations today, I'd say the behavior with `sp-show-pair-from-inside` disabled is fixed, and the behavior with `sp-show-pair-from-inside` is slightly broken.

Here's what I observe as of today, with `$({})$` in a latex buffer.

With `sp-show-pair-from-inside` set to nil, depending on the value of `point`:

1. dollars
2. parentheses
3. braces
4. nothing
5. braces
6. parentheses
7. dollars

With `sp-show-pair-from-inside` set to t:

1. dollars
2. parentheses
3. parentheses
4. braces
5. braces 
6. parentheses
7. dollars

According to the variable docstring, I understand that the expected behavior with `sp-show-pair-from-inside` enabled would be the first behavior, shifted to the "inside":

1. nothing or dollars
2. dollars (the point is after the opening dollar)
3. parenthese
4. braces 
5. parentheses (the point is after the closing brace, but before the closing parenthesis)
6. dollars (idem)
7. nothing or dollars

Incidentally, iirc I was looking for a way to highlight the pair just *after* the point (no matter if we are looking at the opening or closing part of the pair), so that it always highlights the pair *under* the cursor (with a non-bar cursor). The current behavior of `sp-show-pair-from-inside` comes close to achieving a similar behavior...

Fuco1:
[...]

PR #290, opened 2014-01-25
First attempt at fixing bug #289

ThibautVerron:
I hope I didn't break anything. It seems to work in latex, but a similar bug happens with sgml tags, and this doesn't fix it (I suspect it is because the `<` is considered the opening pair of the `</foo>` tag).

If `sp-get-thing` is a slow function, then this "fix" is also highly non-optimized. There is probably a better way to check that we are at the beginning of a sexp anyway.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -1339,7 +1339,8 @@ beginning."
   (memq this-original-command '(self-insert-command
                                 org-self-insert-command
                                 orgtbl-self-insert-command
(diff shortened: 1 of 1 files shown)


Fuco1:
I think 289 is now fixed by tweaking the highlighting priorities.  I will close this PR since it is probably no longer needed.  Thank you for your contribution and the discussion at #289 @ThibautVerron !

PR #394, opened 2014-12-10
Replace delete-backward-char with delete-char

markus1189:
`delete-backward-char` is for interactive use ;)

[24pullrequests](http://24pullrequests.com)

Best,
Markus


Fuco1:
I've commited this on master myself because your patch was branched from almost a year old commit, and rebasing was too much work to change just one line.

Thanks for reporting the warning!

PR #402, opened 2015-01-04
Add optional subword-mode support

johnmastro:
If `sp-use-subword` (a new defcustom) is non-nil and `subword-mode` is active, `sp-kill-word` and `sp-backward-kill-word` will only kill "subwords".

To preserve existing behavior, `sp-use-subword` is nil by default.

Is this a behavior you'd be interested in adding? Does this look like a reasonable way to go about it?

Thanks!

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -1144,6 +1144,12 @@ mute. Integers specify the maximum width."
                  (integer :tag "Max width"))
   :group 'smartparens)
 
+(defcustom sp-use-subword nil
+  "If non-nill, `sp-kill-word' and `sp-backward-kill-word' only
+  kill \"subwords\" when `subword-mode' is active."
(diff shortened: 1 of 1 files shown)


Fuco1:
Very interested, in fact, I have had this feature in my notes for at least half a year now :P I'm running to catch the train now, will review on the way.

Thanks for getting involved!

johnmastro:
My pleasure. By the way, I struggled with whether to keep the names and docstrings short and sweet (since they're so simple) or to give them more explanatory names (e.g. `sp--word-or-subword-forward` or something) and/or more explanatory docstrings. I'd definitely be happy to make a change if you have a preference one way or another.

Fuco1:
I like it the way it is. Merging.

PR #403, opened 2015-01-06
Don't reindent after kill for some modes.

sandinmyjoints:
Indentation in [coffee-mode](https://github.com/defunkt/coffee-mode) is not very smart, it just cycles through possible indentations. So automatically reindenting after a kill mostly results in a correctly indented line becoming incorrectly indented.

This is an attempt to fix that behavior. If there is another, better way to do it, please let me know!

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -440,7 +440,9 @@ Symbol is defined as a chunk of text recognized by
                                              coffee-mode
                                              js2-mode
                                              )
-  "List of modes that should not reindent after kill.")
+  "List of modes that should not reindent after kill."
+  :type '(repeat symbol)
+  :group 'smartparens)
 
 (defvar sp--html-modes '(
                          sgml-mode
--- a/smartparens.el
+++ b/smartparens.el
@@ -436,6 +436,11 @@ Symbol is defined as a chunk of text recognized by
                          common-lisp-mode)
   "List of Lisp modes.")
 
+(defvar sp--no-reindent-after-kill-modes '(
+                                           coffee-mode
+                                           js2-mode)
+  "List of modes that should not reindent after kill.")
+
 (defvar sp--html-modes '(
                          sgml-mode
                          html-mode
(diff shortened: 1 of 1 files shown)


Fuco1:
I'm 80% sure there already is a list like this (or some other solution) as we had the same issue with haskell mode and some others, and we solved it "somehow".  I'll have to look.

sandinmyjoints:
OK. I'd definitely be interested in that. And having used this for a bit more, I think I would update it to not reindent after some other actions such as slurping and barfing as well. 

Bigger picture, smartparens's config system for pair behavior is very flexible -- perhaps similar config around sexp actions would also be helpful (but also more work, I'm sure)?

Fuco1:
I think for now a good solution would be to add hooks before/after the kill and let users install whatever they want, including a reindent action... this action can be set-up automatically by SP for some common modes (it's fairly safe in lisp, java etc...)

pronobis:
That would be great for python too! Will that PR be merged?

jgarvin:
I haven't tried it yet, but I thought that in the latest version of emacs
Python mode was getting updated to have much smarter indentation, with
those changes leaving it on might be desirable.

pronobis:
It might be smarter, but I basically consider deleting a symbol and indentation separate operations. Even if something is misindented (or temporarily misindented since I am inserting code in between existing lines), I would prefer it to stay that way while I am making edits.

mdorman:
Also, Haskell mode has this same issue, so I'm not sure there's some other solution somewhere in the codebase...

Fuco1:
Yes, it seems you are right.

Let's merge this then.

smartparens.el
```diff
@@ -436,6 +436,9 @@ Symbol is defined as a chunk of text recognized by
                          common-lisp-mode)
   "List of Lisp modes.")
 
+(defvar sp--no-reindent-after-kill-modes '(coffee-mode js2-mode)
```
Fuco1: Could you put each mode on its own line, see for example `sp--html-modes` variable.

Also turn it into a defcustom and change the prefix to `sp-` not `sp--`, this option should be user-customizable.

sandinmyjoints:
@Fuco1 Updated per your comments.

smartparens.el
```diff
@@ -436,6 +436,12 @@ Symbol is defined as a chunk of text recognized by
                          common-lisp-mode)
   "List of Lisp modes.")
 
+(defcustom sp-no-reindent-after-kill-modes '(
+                                             coffee-mode
+                                             js2-mode
+                                             )
+  "List of modes that should not reindent after kill.")
```
Fuco1: Thanks! One last thing, put a `:type '(repeat symbol)` and `:group 'smartparens` here so that emacs can construct the customize UI properly---see for example the variable `sp-ignore-modes-list`.

sandinmyjoints:
OK, updated again @Fuco1 . Thanks for your patience. I don’t do much Emacs Lisp coding!

Fuco1:
Awesome, thanks! Merging.

PR #406, opened 2015-01-08
Improve sp-region-ok-p

### LINKED ISSUE #405: sp-region-ok-p returns nil on valid regions an truthy on invalid
Hey, I was hoping to build upon `sp-region-ok-p` but it doesn't appear to be as general as I'd hoped.  Is this something you'd be willing to fix?

Examples of stuff that seems weird:

Valid that are considered deemed invalid:
`,` or `.`

I'd expect all text that doesn't contain any sort of pairs to be considered valid.

Invalid that is deemed valid: 

`foo"`

but `foo)` or `foo]` etc are all deemed invalid.  Why is `"` treated differently?

expez:
The following edge-cases have been fixed:

* Regions that were trivially ok because they didn't contain any opening
  or closing pairs.
* Regions containing an unbalanced string.

Closes #405

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -7512,15 +7512,39 @@ of the point."
         (indent-sexp))
       (sp--back-to-indentation column indentation))))
 
+
+(defun sp--unbalanced-string-after-point-p ()
+  (push major-mode sp-navigate-consider-stringlike-sexp)
+  (save-excursion
+    (unwind-protect
+        (let ((str (ignore-errors (sp-get-string))))
+          (when str
+            (goto-char (plist-get str :beg))
+            (sp-down-sexp)
+            (if (= (point) (save-excursion (sp-up-sexp) (point)))
+                t
+              nil)))
+      (progn (pop sp-navigate-consider-stringlike-sexp) nil))))
+
 (defun sp-region-ok-p (start end)
   (save-excursion
     (save-restriction
       (narrow-to-region start end)
       (goto-char (point-min))
-      (let ((r t))
-        (while (and r (not (eobp)))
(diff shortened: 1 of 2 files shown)


Fuco1:
Thanks a lot for "getting your hands dirty" :) I'll try to review this and merge today evening, or over the weekend, depending on how tired I return from home.

expez:
I've kept pushing new versions of this, but I can't get everything right.  I've added some tests this time as well.

Fuco1:
What cases are problematic?

expez:
Smartparens is so forgiving compared to paredit, e.g. it will let you do `sp-down-sexp` on `foo"` or `foo(` where I'd like it to to error out.  So either I get the regular pairs right and strings go to hell, or I get strings right and the other pairs blow up.

Fuco1:
For me, `sp-down-sexp` doesn't work on parens but works on strings in your above example. Strings are special and damn annoying :/

If you'll work on it more, push whatever you have, I'll also try to get some time and look over some cases. The parser could use a phase-out and a rewrite... but it's so much :(

expez:
Right, it seems to work now, but it's looking like it's doing a lot of work.  I'm calling this function in a tight loop trying to shrink region to provide dwim functionality in `evil-smartparens`.  Hopefully it'll still be plenty fast enough.

Fuco1:
Looks good. I'm going to merge this and make a note to possibly optimize later. If you find it too slow, open a new issue so we have a reminder (and provide some scenarios).

Thanks again.

expez:
:+1:

PR #408, opened 2015-01-11
Add inf-clojure-mode to sp--lisp-modes

bbatsov:
See https://github.com/clojure-emacs/inf-clojure/blob/master/inf-clojure.el


Fuco1:
Thanks. I'll merge all three of these when I get back home.

PR #409, opened 2015-01-11
Remote nrepl-repl-mode from sp--lisp-modes

bbatsov:
This mode was renamed to cider-repl-mode about 1 year ago, so I think we can safely remove it.


bbatsov:
Damn, I made a typo in the commit message...

Fuco1:
Merged via rebase.

PR #414, opened 2015-01-12
improve-sp-region-ok for trailing ws and comment

expez:
Regions with trailing ws or comments would not be validated as OK.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -7427,10 +7427,20 @@ of the point."
               nil)))
       (progn (pop sp-navigate-consider-stringlike-sexp) nil))))
 
+;;; Taken from subr-x.el
+(defsubst string-trim-right (string)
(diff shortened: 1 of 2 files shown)


Fuco1:
If you are going to work on this, it would be simpler to give you push access. What do you think?

expez:
sure

expez:
Although I would prefer that you slept less--look into [polyphasic sleep](http://en.wikipedia.org/wiki/Polyphasic_sleep)-- and fixed the bugs yourself :P

Fuco1:
I tried that while in college, it doesn't work (at least for me).

Other than getting shittons of money, having a job kind of sucks :/

I've added you to the project, so you can push this to master.  I don't worry much if something breaks, people can use the stables/marmalade. Just don't `--force` push :)

expez:
:+1:

PR #450, opened 2015-04-05
 support python.el as well as python-mode.el

djr7C4:
Note: I didn't add a sp--python-modes variable since python.el and python-mode.el both provide python-mode, so that wouldn't make sense.


expez:
Thanks! :+1:

PR #452, opened 2015-04-08
Where sp-kil-hybrid-sexp wasn't killing whitespace and it is now fixed. 

mpontus:
This change makes `sp-kill-hybrid-sexp` to kill a region of variable length depending on whitespace and value of `sp-hybrid-kill-excessive-whitespace'. 

By attracting both kill-region and delete-region into single command it was made possible to make this data available in kill ring without losses.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -1008,10 +1008,14 @@ by specifying its :suffix property."
 
 ;; hybrid lines
 (defcustom sp-hybrid-kill-excessive-whitespace nil
-  "If non-nil, `sp-kill-hybrid-sexp' will kill all whitespace up
-until next hybrid sexp if the point is at the end of line or on a
-blank line."
-  :type 'boolean
+  "If non-nil, `sp-kill-hybrid-sexp' will delete all whitespace
+up until next hybrid sexp if the point is at the end of line or
+on a blank line. When argument is 'kill whitespace will meld to
+deleted sexp in kill ring."
+  :type '(choice
+          (const :tag "On" t)
+          (const :tag "Kill" kill)
+          (const :tag "Off" nil))
   :group 'smartparens)
 
 (defcustom sp-hybrid-kill-entire-symbol nil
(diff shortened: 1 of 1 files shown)


Fuco1:
I think the idea here actually was to not have the whitespace in the kill ring, because more often than not you don't care about it.  Maybe there should be another switch to decide if we want to append the whitespace to the killed text or not.

myeffort:
Well could be. Let's option then.

smartparens.el
```diff
@@ -5189,15 +5193,16 @@ Examples:
...
-            (kill-region (point) (min (point-max) (if (looking-at "[ \t]*$") (1+ :end-suf) :end-suf)))
-            (when sp-hybrid-kill-excessive-whitespace
-              (cond
-               ((sp-point-in-blank-line)
-                (while (and (not (eobp))
-                            (sp-point-in-blank-line))
-                  (delete-region (line-beginning-position) (min (point-max) (1+ (line-end-position))))))
-               ((looking-at "[ \t]*$")
-                (delete-blank-lines)))))))
+            (let ((end (min (point-max)(if (looking-at "[ \t]*$")
```
Fuco1: Add a space after the `(point-max)` form and format the `if` form according to elisp rules (this looks like scheme)

smartparens.el
```diff
@@ -1008,10 +1008,14 @@ by specifying its :suffix property."
 
 ;; hybrid lines
 (defcustom sp-hybrid-kill-excessive-whitespace nil
-  "If non-nil, `sp-kill-hybrid-sexp' will kill all whitespace up
-until next hybrid sexp if the point is at the end of line or on a
-blank line."
-  :type 'boolean
+  "If non-nil, `sp-kill-hybrid-sexp' will delete all whitespace
```
Fuco1: This is no longer true now, non-nil doesn't mean that, only `t`.  Add a description for the third state too (when the option is `nil`)

myeffort: Killing is deletion, is it not? Besides anything but 'kill will delete and not kil.

Fuco1:
All right.  I'm going through older issues/PRs, trying to get things moving again.

I've review the code and it looks good.  We should probably get some tests going too and then this is good to merge.

Fuco1:
Commited in 444bfb0

PR #453, opened 2015-04-08
sp-clone-sexp will no longer dash ahead.

myeffort:
This function had issues with relative cursor position which made it distance itself further and further from original context. This is fixed version, as well as taking some liberties in deciding what most useful context would be. 
The behavior of `sp-get-sexp` responsible for choosing its targets reflects on this command in a way that delimits its usefulness. Especially in the root context, where `sp-get-enclosing-sexp` doesn't shine with efficiency either.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -5135,11 +5135,15 @@ t.  All the special prefix arguments work the same way."
 
 (defun sp-clone-sexp ()
   (interactive)
-  (-when-let (ok (sp-get-thing))
+  (-when-let (ok (or (sp-get-enclosing-sexp)
+                     (sp-get-sexp)))
     (sp-get ok
-      (goto-char :end-suf)
-      (sp-newline)
-      (insert (buffer-substring-no-properties :beg-prf :end-suf)))))
+      (save-excursion
+        (undo-boundary)
+        (goto-char :beg-prf)
+        (insert-buffer-substring-no-properties
+         (current-buffer) :beg-prf :end-suf)
+        (newline-and-indent)))))
 
 (defun sp-kill-hybrid-sexp (arg)
   "Kill a line as if with `kill-line', but respecting delimiters.


Fuco1:
I'm not sure about using `sp-get-enclosing-sexp`.  Why would I want to clone the enclosing sexp and not the one after the point?

This should probably not use `sp-get-sexp` directly at all and rely on `sp--next-thing-selection` with symbols as sexps turned off.

myeffort:
May I ask in circumstances do you use clone function most often?

Fuco1:
I don't use it at all.

myeffort:
And I don't find enough opportunity to watch others' workflow than I would like. 
So I can only speak for myself, I found clone functions to be useful when writing semi-repetitive forms such as define-key, series of setq's, let assignments, argument to `easy-mm-define-keymap` and so on. Clearly all of them are surrounded in parens.

Sometimes I'd make a clone of top level form to experiment on it. Currently this is not as easy to do as I would like. So if I want to clone a symbol, it's the least prevalent situation, yet still admittedly useful with setq's and setq-default's. If I could have them I would, but now I chose most useful.

Ideally I would like to clone a symbol if cursor belongs to it, and whenever its not, it should make clone of surrounding form. To clarify, as you mentioned cloning sexp after point, I would prefer in such cases to clone surrounding sexp.

If you would give it a try, take advantage of static positioning, and you will see why it is more productive to clone sexp which you can already affect, and not something by the side. To save you a bit of trouble, I have it bound to M-c.

Fuco1:
These are all valid reasons and I understand how it can be useful.  But I don't want to break the backward API for the people who use it now (at least one person does it, the one who submited the original PR for this function)

How about we add `sp-clone-sexp-dwim` which would behave in a more context-aware way?

myeffort:
Should there be no attempt then to change the way how cursor moves as well?
I find the function introduced by original PR to be utterly broken and
accepting in its form to be a mistake.

Fuco1:
Very well... if someone will complain, they can copy the old code over.

Fuco1:
I played with your version a bit and it is indeed very pleasant to use. I like it.

Thanks for convincing me over :)

PR #455, opened 2015-04-11
Add hooks to limit searching within bounds.

### LINKED ISSUE #451: Double quote behaviour is wrong in slime-repl
In the slime-repl, double quotes do not work at all.

The first time you enter a double quote it inserts a single quote, but the second time, it insertst two:

1:   `(format nil |)`
2 :  `(format nil "|)` After pressing "
3:   `(format nil "foo|)` 
4:   `(format nil "foo"|")` after pressing " again

This only seems to happen inside of an s-expre (a string as the top level expression works fine).
It is extremely inconvenient because it leaves me with an odd number of quotes, and I have to either use C-q " to insert a single quote or C-0 &lt;DEL&gt; to delete one of the quotes.

tmccombs:
Fixes #451

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -156,6 +156,21 @@ better orientation."
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Variables
 
+(defvar-local sp-forward-bound-fn nil
+  "Function to restrict the forward search")
+
+(defvar-local sp-backward-bound-fn nil
+  "Function to restrict the backward search")
+
+(defun sp--get-forward-bound ()
+  "Get the bound to limit the forward search for looking for pairs."
+  (and sp-forward-bound-fn (funcall sp-forward-bound-fn)))
+
+(defun sp--get-backward-bound ()
+  "Get the bound to limit the backward search for looking for pairs."
+  (and sp-backward-bound-fn (funcall sp-backward-bound-fn)))
+
+
 ;;;###autoload
 (defvar sp-keymap (make-sparse-keymap)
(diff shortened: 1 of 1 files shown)


Fuco1:
Wow, this solution is genius :) Much better than what I had in mind, and nicely localized.

So, what exactly you do to set the bounds?  We might want to add a section to the documentation to explain how this works.

smartparens.el
```diff
@@ -156,6 +156,21 @@ better orientation."
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Variables
 
+(defvar sp-forward-bound-fn nil
+  "Function to restrict the forward search")
+
+(defvar sp-backward-bound-fn nil
+  "Function to restrict the backward search")
```
Fuco1: Shouldn't we make these buffer local? So that each mode can set different values? Or will be expect the user to call `make-local-variable`?  If so, it needs to be documented because it's quite simple to forget that, then weird things might happen globally.

But it is true that this will stay nil most of the time and so adding a local version to every buffer is quite needless.

tmccombs:
I changed it to use buffer-local variables. 

Here is what I have to set it for the slime repl.

```lisp
(add-hook 'sime-repl-mode (lambda ()
                            (setq sp-backward-bound-fn (lambda ()
                                                         slime-repl-input-start-mark))))
```

expez:
This is going to be lovely for the `CIDER` repl as well!

Fuco1:
Have you pushed the new commit?

tmccombs:
Yes.  I amended the original commit.

expez:
Thanks! :+1:

PR #459, opened 2015-04-23
Fix point at wrong side of bound error

expez:
The functions `sp--search-{forward,backward}-regexp` appear to have had
their bounds reversed.

Does the seem right to you, @tmccombs?


tmccombs:
Yep,  my bad.

Fuco1:
Ha, thanks for catching this :)

PR #460, opened 2015-04-28
First stab at matching haskell comment blocks

### LINKED ISSUE #424: Suggesting a pair for haskell-mode
For a long time I had an issue with smartparens where if my point was on a LANGUAGE pragma, the minibuffer would constantly have the following error "Search failed. This means there is unmatched expression somewhere or we are at the beginning/end of file".

You can reproduce this by opening a buffer in Haskell mode and start it like this:

```haskell
{-# LANGUAGE OverloadedStrings #-}
```

The thing that fixed it was:

```elisp
(sp-local-pair 'haskell-mode "{-#" "#-}")
```

I advocate adding this to smartparens itself so haskell developers don't have to see this weird error message and research to come to the same conclusion. I presume that this would not have any negative impact for users who don't use haskell as it would never get activated.

Thanks

MichaelXavier:
Ok there's my first stab at it. Didn't know what you wanted to do with regards to the file header so I left that blank. This should resolve #424

### DIFF
--- a/smartparens-config.el
+++ b/smartparens-config.el
@@ -100,6 +100,8 @@
 (eval-after-load "lua-mode"      '(require 'smartparens-lua))
 (eval-after-load "ruby-mode"     '(require 'smartparens-ruby))
 (eval-after-load "enh-ruby-mode" '(require 'smartparens-ruby))
+(eval-after-load "haskell-mode"     '(require 'smartparens-haskell))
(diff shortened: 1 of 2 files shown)


Fuco1:
Thanks.

You can copy the header from `smartparens-latex` or `ruby`, put your name there, update the dates `(c) 2015` etc.

MichaelXavier:
Done!

Fuco1:
Awesome :) Thanks

PR #443, opened 2015-03-05
Allow shift selection with forward/backward sexps

bobbysmith007:
re Fuco1/smartparens#360

PR #474, opened 2015-06-02
Enable CUA shift selection with smartparens motions

### LINKED ISSUE #360: Selecting sexps one-by-one?
Currently smartparens seems to ignore Shift when doing `sp-backward-sexp` and `sp-forward-sexp`, which makes selecting-by-sexps impossible; `sp-select-next-thing` doesn't work either, because it replaces selection instead of incrementing it. Basically, what I want is for N invocations of `sp-select-next-thing` to be equivalent to `sp-select-next-thing N` (and for `sp-select-prev-thing` to undo 1 invocation of `sp-select-next-thing`). Would it be possible to implement something like this in smartparens?

The attached code demonstates the behavior I'm describing (it's horribly inefficient, being the first piece of code I've ever written in Emacs Lisp – sorry!):

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ elisp
(define-key global-map (kbd "C-S-<left>" ) 'sp-backward-sexp-select)
(define-key global-map (kbd "C-S-<right>") 'sp-forward-sexp-select)

(defvar sp-selection-level 0)
(defvar sp-selection-point nil)
(defun sp-add-and-reselect (arg)
  (interactive "p")
  ; Initialise.
  (unless (use-region-p)
    (setq sp-selection-level 0)
    (setq sp-selection-point nil))
  (unless sp-selection-point
    (setq sp-selection-point (point)))
  ; Clear selection – sp-select-next-thing doesn't work well otherwise.
  (set-mark sp-selection-point)
  (goto-char sp-selection-point)
  ; Increment level.
  (setq sp-selection-level (+ sp-selection-level arg))
  (unless (eq 0 sp-selection-level)
    (sp-select-next-thing sp-selection-level sp-selection-point)
    ; If level is negative/positive, point has to be before/after mark; if
    ; it's not the case, do the swap.
    (unless (or (and (> 0 sp-selection-level) (< (point) (mark)))
                (and (< 0 sp-selection-level) (> (point) (mark))))
      (exchange-point-and-mark))))

(defun sp-backward-sexp-select ()
  (interactive)
  (sp-add-and-reselect -1))
(defun sp-forward-sexp-select ()
  (interactive)
  (sp-add-and-reselect 1))
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

PythonNut:
@Fuco1 

This should be pretty self explanatory. Should fix #360. Should supersede #443.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -4639,7 +4639,7 @@ Examples: (prefix arg in comment)
   (|foo bar baz)   -> (foo bar| baz) ;; 2
 
   (foo (bar baz|)) -> (foo (bar baz)|)"
-  (interactive "p")
+  (interactive "^p")
   (setq arg (or arg 1))
   (if (< arg 0)
       (sp-backward-sexp (- arg))
@@ -4671,7 +4671,7 @@ Examples: (prefix arg in comment)
   (foo bar| baz)   -> (|foo bar baz) ;; 2
 
   (|(foo bar) baz) -> ((|foo bar) baz)"
-  (interactive "p")
+  (interactive "^p")
   (setq arg (or arg 1))
   (if (< arg 0)
       (sp-forward-sexp (- arg))
@@ -4699,7 +4699,7 @@ Examples:
   ((foo) |bar (baz quux)) -> ((foo) bar |(baz quux))
 
   ((foo) bar |(baz quux)) -> |((foo) bar (baz quux))"
-  (interactive "p")
+  (interactive "^p")
   (setq arg (or arg 1))
   (if (> arg 0)
       (if (= arg 1)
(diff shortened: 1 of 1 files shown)


Fuco1:
I think I tried this once and it didn't work. Do you use cua? Does it work for you?

Fuco1:
Also the person in other PR (which sadly I missed :( sorry for that) adds the cua property nonsense. Is that not also necessary?

PythonNut:
Actually, it is, but I added it to my personal config years ago and forgot about it. Give me a moment.

PythonNut:
There. 

> Do you use cua? Does it work for you?

Yes I do, and yes it dos.

Fuco1:
Okey looks good :)  Thanks.

Issue #475, opened 2015-06-03
Let's have a menu


andreas-roehler:
PR considered.

Fuco1:
What would it do?

Fuco1:
This is solved in #476, closing.

PR #476, opened 2015-06-03
#475 fixed, Lets have a menu

andreas-roehler:


Fuco1:
Does it have to duplicate the docstrings? That's a bit unfortunate, then one would need to keep it in sync (and it bloats the file a lot!).  I don't know how easymenu works but this seems suboptimal :/

andreas-roehler:
Am 03.06.2015 um 22:18 schrieb Matus Goljer:
>
> Does it have to duplicate the docstrings?
>

No. Menu seems interesting for beginners mostly.
For them some info WRT to command was delivered.

Still want  the docstrings deleted?

Fuco1:
I get it's more convenient to see it there right away than doing `C-h f` but I really fear this will get out of sync very soon.  Maybe there's a way to retrive docstring of a function programatically and insert that into the menu?

andreas-roehler:
Am 04.06.2015 um 19:58 schrieb Matus Goljer:
>
> I get it's more convenient to see it there right away than doing |C-h 
> f| but I really fear this will get out of sync very soon. Maybe 
> there's a way to retrive docstring of a function programatically and 
> insert that into the menu?
>
> —
>

It would be possible to refresh the menu like cheat-sheat is printed out.

That works to the extend, as naming of the commands is regular.

Pushed the code which generated basics of sp-menu, see at end of file:

https://github.com/andreas-roehler/werkstatt/blob/master/devel/ar-create-menu.el

Fuco1:
Sorry I left you waiting for so long.  I am now getting a bit of time, I promise I will review this by the end of next week.  If you made any changes, please push it, and also rebase against the master. Thanks!

PR #494, opened 2015-08-05
Add custom indentation function.

rpglover64:
Also add supporting code.

Attempts to address Fuco1/smartparens/issues/493.


Fuco1:
This now conflicts with #522.  If you want you can base a new patch on that, or I will do it myself (I'm sorry for not merging this sooner, now rendering your work incompatible).  Adding the customize and another check to the `sp--indent-region` function should work.

PR #500, opened 2015-09-02
Always show matching paren over cursor in `evil-mode`

ralesi:
When in normal and visual states in `evil`, it doesn't make sense to respect `sp-show-pair-from-inside`.  Change shows matches in normal and visual states matching the pair under the cursor.  This also ensures `evil-jump-item` or `%` will always move to the pair show-paren is highlighting.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -1248,6 +1248,10 @@ insert the modes."
   "Checks to see if the current `evil-state' is in normal mode."
   (and (fboundp 'evil-normal-state-p) (evil-normal-state-p)))
 
+(defun sp--evil-motion-state-p ()
+  "Checks to see if the current `evil-state' is in motion mode."
+  (and (fboundp 'evil-motion-state-p) (evil-motion-state-p)))
+
 (defun sp--evil-visual-state-p ()
   "Checks to see if the current `evil-state' is in visual mode."
   (and (fboundp 'evil-visual-state-p) (evil-visual-state-p)))
@@ -7915,10 +7919,15 @@ support custom pairs."
            ;; if we are in a situation "()|", we should highlight the
            ;; regular pair and not the string pair "from inside"
            ((and (not (sp--evil-normal-state-p))
+                 (not (sp--evil-motion-state-p))
                  (not (sp--evil-visual-state-p))
                  (sp--looking-back (if sp-show-pair-from-inside allowed closing)))
             (scan-and-place-overlays (match-string 0) :back))
-           ((or (sp--looking-at (if sp-show-pair-from-inside allowed opening))
+           ((or (and (or (sp--evil-normal-state-p)
+                         (sp--evil-motion-state-p)
+                         (sp--evil-visual-state-p))
+                     (sp--looking-at (sp--get-allowed-regexp)))
+                (sp--looking-at (if sp-show-pair-from-inside allowed opening))
                 (and (memq major-mode sp-navigate-consider-stringlike-sexp)
                      (looking-at (sp--get-stringlike-regexp)))
                 (and (memq major-mode sp-navigate-consider-sgml-tags)


Fuco1:
Thanks for the initiative.

I'm not an evil user myself, do you think this is universally accepted to be the default?  If you know some other evil users, maybe they can also express their opinion here.  Thanks!

ralesi:
These changes are also consistent with how `show-paren-mode` functions for evil normal and visual states.

sooheon:
This is definitely the way vim does it, but given the number of lisp commands in emacs which operate on previous sexp, highlighting the closing parens just before point is desirable for some people. Because sp-show-pair-from-inside exists, what about adding another possible value to it (`'evil`) to get the behaviour you want, rather than making this behaviour automatic on detection of normal/visual state? It's a grey area anyway, so more customisability trumps accurate vim emulation, imo.

ralesi:
@sooheon, I can't see anywhere in the source that sp-show-pair-from-inside is used outside of the show-parens function.  Can you give me an example of another major mode using sp-prev-sexp that will help me figure out how it would be different.  At least in the case of elisp, it is obvious we want the matching pair the cursor is on.

sooheon:
Ah I don't mean that it is used in source, but that a lot of evaluation in emacs works on the *previous* s-exp (`eval-last-sexp`, `cider-eval-last-sexp`), so while navigating code, it is useful to see the preceding sexp highlighted when cursor is after a closing paren. Again, I see why you would prefer it to be like vim, but I have my preferences as well, so a choice in the value of `show-pair-from-inside` would be better imo.

ralesi:
@sooheon, thanks for taking the time out to explain this to me.  I understand you have a preference and am trying to understand it.  I am fine with adding another option, but I feel like I am missing the use case when we want the previous - previous sexp.  

In the case of normal / visual mode, in this elisp example, if | | is to surround the position of the cursor, I would expect : 

````
(defun ()
    "blah blah"
    (setq blah (concat blah "blah"))|)|
````

to highlight the paren in front of `defun` in all cases.  Is there a case you would not want this?  Are there cases in which you want to highlight both side of the previous `setq` statement, even if the cursor is not over that paren?

The problem I see with adding another option to `sp-show-pair-from-inside` is that we don't have the ability to cover options for non normal / visual states when the option would be set as `evil`.  In the current PR, when we revert to insert / emacs state, we still use the preferred option of being within / outside the previous sexp.

**Comment by Fuco1** [comment]
When you are not using evil and call for example `eval-last-sexp` and you are "on" the `)` it won't eval the list but the preceeding symbol `(+ 1 2 3|)` would eval to `3` not 6 as it would in evil.  So highlighting the opening paren is inaccurate because emacs is not operating on that sexp (and it's not only eval but also copy/select/move/forward etc).

Emacs and vim have different notion of where cursor is and emacs very consistently follows the rule that if the cursor is *at* a character it is actually after the *previous one*.  In vim this is reversed.

So in your situation

```elisp
(defun ()
    "blah blah"
    (setq blah (concat blah "blah"))|)
```

I indeed *don't* want to see the opening paren at `defun` highlighted... because that's not the sexp I'm working on, it should instead highlight the one at `setq`.

Fuco1:
n.b. from what @sooheon said I'm not sure now if evil indeed does change `eval-last-sexp` or operates the same way as emacs... still the distinction remains.

ralesi:
Here is an example of how show-smartparens works in evil normal and visual state when you perform `eval-last-sexp`.  To me it looks like it is doing exactly what it should be.  When on the trailing paren, it selects the matching pair from the cursor, and evals that pair.  Let me know if this make sense or not or if I am still missing something.  
evil-smartparens

geo7:
Makes sense to me @ralesi :+1:

sooheon:
@ralesi Do you have some advices on eval-last-sexp which give you this behaviour? Or is it this PR that affects those changes? Ideally, in visual/normal state we should highlight paren directly under cursor *and* eval sexp ending with that paren, and in insert state should highlight paren before point and eval sexp ending before point. Am I making sense?

ralesi:
@sooheon,  I dont have any advices on `eval-last-expr`.  The only thing different in the image is this PR.  This PR should give you your desired behavior.  I don't touch insert-state at all.

sooheon:
@ralesi I applied your commit to my smartparens.el, and in general I agree with the changes. In insert mode, the highlighting behaves exactly as in emacs, and in normal/visual, it behaves like vim. One edge case is that at the very end of line, if evil-move-beyond-eol is on, it will highlight the last paren before point even in normal or visual state. For consistency, this should be disabled.

Regarding the `eval-last-sexp`, are you sure you don't have any other packages which advise that function? It doesn't work the same way for me, and it always evals the sexp ending *before* point.

ralesi:
@sooheon, I get this behavior without any advices on `eval-last-sexp`.  I have found though that the `evil-move-beyond-eol` is what changes the way evil works by modifying `preceding-sexp`, `pp-last-sexp`.  See evil-integration.el.  When option is set to nil, you get the behavior as shown above.

justmytwospence:
Where did we land on this? I'd also like to see some way to make show-smartparens-mode act like vim, whatever others think the best way to manifest that may be.

smartparens.el
```diff
@@ -7829,7 +7829,10 @@ support custom pairs."
                  (not (sp--evil-visual-state-p))
                  (sp--looking-back (if sp-show-pair-from-inside allowed closing)))
             (scan-and-place-overlays (match-string 0) :back))
-           ((or (sp--looking-at (if sp-show-pair-from-inside allowed opening))
+           ((or (and (or (sp--evil-normal-state-p)
+                         (sp--evil-visual-state-p))
```
expez: motion state?

expez:
With `evil-move-beyond-eol` set to `nil` the suggested patch works as it should and it feels like a solid improvement to me.

Fuco1:
I don't know what those settings do.  @expez will you take care of this PR?  I don't care either way so far as evil users find it reasonable.

expez:
@ralesi Could you amend the commit to make this happen for motion state as well?  Once that's done I think this is merge-worthy.

ralesi:
I will address the motion state changes tonight and rebase on current HEAD.   Any other changes beaides that?

ralesi:
I updated the commit, for some reason Travis CI is failing the build now though.  I can't tell from the build log what went wrong.

expez:
> I can't tell from the build log what went wrong. 

There are some failures on `master` too, with old versions of emacs.

expez:
Thanks for the patch @ralesi!

Fuco1:
The problem is that Cask can't figure out what to do on older versions when optional dependencies aren't met.  I've opened an issue at their repo and we're trying to come to a solution.

Thanks for the patch.

PR #504, opened 2015-09-17
Add special insert commands for quack

### LINKED ISSUE #502: No matching parens with geiser/quack
When using geiser/quack, `(` gets bound to
`quack-insert-opening-paren` which inserts the `"("` but smartparens
doesn't insert the matching `")"`.

vikraman:
These are used in quack for parentheses and brackets.

Fixes #502


Fuco1:
Thanks :)

PR #509, opened 2015-10-03
Add special insert commands for racket-mode

vikraman:
These are used in racket-mode for inserting parens, brackets and braces.


Fuco1:
What about the opening variants, are those not used?

vikraman:
No, those are not used.

Fuco1:
Thanks.

Issue #493, opened 2015-08-05
Add customize option to *not* indent region after e.g. slurping in certain modes


rpglover64:
Certain modes have a broken `indent-region` (c.f. haskell/haskell-mode#796), and smartparens would be more useful it it didn't try to reindent after modifying the document in these cases.

Fuco1:
See this: https://github.com/Fuco1/smartparens/pull/403

Maybe we should extend that setting to all the "post-command automatic attempts to fix indent".

If someone cares to take it up, I wouldn't mind.  This is a low priority task for me right now, unfortunatelly I barely have time to work on the "important" issues anymore :/

rpglover64:
I noticed that and started hacking something up, but I wasn't sure which commands to do reindent for and which ones not to.

Fuco1:
Well, if the indent breaks thing then I guess it makes sense to either have it or not at all.

But I think a better approach would be not to call reindents at all from the body directly but add it as a hook action.  Then users can add or remove the hook as they like... and we can remove it in the `smartparens-foo.el` config files for the languages where it makes no sense by default.  Most interactive functions should raise an event which can be catched with the pre/post handlers, and if they don't that functionality should be added. (see `sp--run-hook-with-args` in various commands).

Because this global approach is not going to give us too much control.  It's probably a good temporary solution though.

Fuco1:
This is handled in #494 and #513

rpglover64:
It's not fully solved yet... for example, slurping and barfing still cause the indentation issue.

PR #513, opened 2015-10-16
sp-splice-sexp: also avoid indenting here on certain modes

da-x:
Until #493 is resolved, this should make some Haskellers happy.

---

Often in refactoring we remove parentheses in Haskell, e.g. in this case using `sp-splice-sexp`, but haskell-mode indentation suffers from the same problem for that we have had in `sp-kill-sexp`. For example:

```haskell
import Control.Monad
import System.Directory
import System.FilePath

main = do
    content <- getDirectoryContents "."
    forM_ content $ \filename -> do
        let filenameInDir = ("." </> filename)
        print filenameInDir
```

Before this change, `sp-splice-sexp` inside the parentheses would result in:

```haskell
    forM_ content $ \filename -> do
         let filenameInDir = "." </> filename
        print filenameInDir
```

Which breaks compilation. It should be:

```haskell
    forM_ content $ \filename -> do
        let filenameInDir = "." </> filename
        print filenameInDir
```

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -6511,8 +6511,9 @@ represent a valid object in a buffer!"
           (let ((b (bounds-of-thing-at-point 'line)))
             (delete-region (car b) (cdr b))))
         (setq indent-from (point)))
-      (sp--keep-indentation
-        (indent-region indent-from indent-to)))))
+      (unless (memq major-mode sp-no-reindent-after-kill-modes)
+        (sp--keep-indentation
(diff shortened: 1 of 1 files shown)


da-x:
The CI failure seems unrelated.

Fuco1:
Well, this is an easy fix so why not.  Hopefully some day we'll get a better solution.

Thanks!

PR #520, opened 2015-10-27
Use `comint-last-prompt' in smartparens-haskell

marsam:
`comint-last-prompt` has superseed `comint-last-prompt-overlay` in recent Emacs versions.


Fuco1:
Thanks, this looks great :)

PR #522, opened 2015-10-28
Don't indent if aggressive-indent-mode is enabled

tsdh:
aggressive-indent-mode [1] already provides support for re-indentation
after buffer changes, and it does it more sophisticated than just
calling indent-region (and additionally, it won't issue an annoying
"Indenting region..." message).  Therefore, if aggressive-indent-mode is
enabled in the current buffer, don't indent the region but let the job
be done by that.

[1] https://github.com/Malabarba/aggressive-indent-mode

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -1233,7 +1233,9 @@ mute. Integers specify the maximum width."
 (defun sp--indent-region (start end &optional column)
   "Call `indent-region' unless `aggressive-indent-mode' is enabled."
   (unless (bound-and-true-p aggressive-indent-mode)
-    (indent-region start end column)))
+    ;; Don't issue "Indenting region..." message.
+    (cl-letf (((symbol-function 'message) #'ignore))
+      (indent-region start end column))))
 
 (defmacro sp-with-modes (arg &rest forms)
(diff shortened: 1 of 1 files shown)


manuel-uberti:
:+1: yes please!

smartparens.el
```diff
@@ -6825,8 +6832,8 @@ We want to move the `while' before the `let'.
                                      :end)))
            (inner-raise (sp-get enc (delete-and-extract-region
                                      :beg-prf
-                                      (save-excursion
-                                        (sp-forward-whitespace)))))
+				     (save-excursion
+				       (sp-forward-whitespace)))))
```
expez: Seems to be a mix of tabs and spaces here.  Tabs are gross.

tsdh: Well, I just use the emacs defaults.  If you want a different style, put a `.dir-locals.el` in the project.

manuel-uberti:
Any news on this one?

Fuco1:
I like this.  Since you abstracted the indent function we can also add an additional check to determine if we want to use SP to reindent there or not (there are problems in haskell and elsewhere with their "weird" indent functions).

PR #525, opened 2015-11-03
Ensure turn-off-smartparens-mode also disables strict mode.

Wilfred:
I've started using smartparens-strict-mode, and it works really well. However, I noticed my calls to `turn-off-smartparens-mode` didn't disable strict mode. This fixes that.

I think this is a reasonable behaviour, let me know what you think.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -715,9 +715,16 @@ in mode's startup-hook etc.) by calling `smartparens-mode'."
 
 ;;;###autoload
 (defun turn-off-smartparens-mode ()
-  "Turn off `smartparens-mode'."
+  "Turn off `smartparens-mode'.
+See also `turn-off-smartparens-trict-mode'."
   (interactive)
   (smartparens-mode -1))
+
+;;;###autoload
+(defun turn-off-smartparens-strict-mode ()
+  "Turn off `smartparens-mode' and `smartparens-strict-mode'."
+  (interactive)
+  (smartparens-strict-mode -1))
 
 ;; insert custom
 (defcustom sp-autoinsert-pair t


Fuco1:
If we disable it like this then turning sp back on would not re-enable strict mode, which is what I would expect.

That is, if I enable strict mode for a buffer I would want `smartparens-mode` to disable it when I disable sp and enable it back when I enable it.  I'm not sure how to solve this reliably.

Wilfred:
You're right, that'd be a little confusing. How about we add a `turn-off-smartparens-strict-mode` function? If users want both switched off, they can call both functions (it'd suit me at least).

I've modified my branch to add this command.

Fuco1:
Is there, then, any reason not to call `(smartparens-strict-mode -1)`.  Or do you want it for hooks? Notice that calling this won't turn off `smartparens-mode` so you would need to call both "turn off" functions.

Wilfred:
Yep, exactly, I'd like it for hooks. I suppose I could call `(smartparens-strict-mode -1)` but we could say the same thing about `turn-off-smartparens-mode`. I like how explicit `turn-off-smartparens-strict-mode` is.

Fuco1:
Maybe SP could have its own extra local variable which would track if the strict mode was on when sp was being disabled and re-enable it according to that.

So the situations will be:
1. User turns both on.  If a setting "track strict mode" is on, toggling SP also toggles strict mode (first time turning on strict mode means it is "on" for the buffer)
2. User turns both on, but has "track strict mode" disabled.  To disable both, they have to turn both off manually/on hook (like you propose)
3. User doesn't use strict mode at all.  Nothing special happens.
4. User toggles strict mode on and then off, the "tracking variable" will set itself to off.  Toggling SP on or off will leave strict mode off.

Since we have 2 modes with 2 states 4 situations should be all, unless I'm missing some sequence

Wilfred:
Hm, is it that complicated? Are there any situations where I'd want `smartparens-strict-mode` but not `smartparens-mode`?

I thought strict mode was just SP with a few extensions. I can't see any situation where I wouldn't want to switch one off without also wanting to switch the other off.

Fuco1:
Right, the issue is more that if you had strict mode before and you toggle both off with just call to `turn-off-smartparens-mode`, calling `turn-on-smartparens-mode` (or just `(smartparens-mode 1)` should also re-enable strict mode.

You are right that it makes little sense having just strict mode on without SP.  The reason it's a minor mode is to have the keymap with the remapping toggle-able.

Fuco1:
The points from the above comment are now implemented.

PR #526, opened 2015-11-03
Handle ' in rust-mode intelligently.

### LINKED ISSUE #512: Don't treat ' as a delimiter in rust-mode
In rust, lifetime specifiers look like this: `'foo`. Whilst it's also possible to use `'` for single characters, e.g. `'h'`, lifetime specifiers are more common.

Wilfred:
Fixes #512.

### DIFF
--- a/Cask
+++ b/Cask
@@ -13,4 +13,5 @@
   (depends-on "shut-up")
   (depends-on "ruby-mode")
   (depends-on "racket-mode")
+  (depends-on "rust-mode")
   (depends-on "auctex"))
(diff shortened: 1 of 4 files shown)


Fuco1:
You can use `sp-test-with-temp-buffer` instead of your own setup function.  It should be understandable from the docstring or you can search the tests for an example.

I will write something about the "test framework" hopefully soon enough :P

Wilfred:
Aha, I didn't know about `sp-test-with-temp-buffer`. I've pushed a new commit, let me know if you'd prefer I squashed it.

Fuco1:
Yes please, squash and then we merge! Thanks.

Wilfred:
Squashed :)

Fuco1:
Bam, thanks!

skeuomorf:
Nice! Thanks @Wilfred and @Fuco1 :)

PR #527, opened 2015-11-03
Don't ignore errors when running :when and :unless functions.

Wilfred:
Otherwise, this can lead to very hard-to-debug errors.

Consider the code:

    (defun always-true (id context)
      t)

    (sp-with-modes '(some-mode)
      (sp-local-pair "[" "]" :unless '(always-true)))

We would expect [ to never be automatically paired. However,
`always-true` only takes 2 arguments, when it should take 3. As a
result, we swallow the error and consider `always-true` to have returned
nil!

This made my debugging tricky, so I believe throwing the error
sooner is the right thing to do here.

### DIFF
--- a/smartparens-config.el
+++ b/smartparens-config.el
@@ -53,16 +53,19 @@
 
 (defun sp-lisp-invalid-hyperlink-p (_1 action _2)
   (when (eq action 'navigate)
-    (or (and (looking-at "\\sw\\|\\s_")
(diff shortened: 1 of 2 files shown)


Fuco1:
Those ignores were added almost 3 years ago to fix a bug we didn't know how else to fix @_@ So I guess it's about time to remove them and address the issue properly.  Thanks.

PR #536, opened 2015-11-16
Correcting docstring that erroneously said prefix.

Wilfred:


Fuco1:
These are annoying to merge, I added you as collaborator :) Feel free to do similar patches on your own, if you need some discussion you can do a pull request (from this repo or otherwise).

You have been active a lot lately so why not simplify it.

PR #538, opened 2015-11-16
When slurping forward, don't include punctuation at end of line.

### LINKED ISSUE #534: sp-slurp-hybrid-sexp should not surround punctuation
Given the JS:

```js
var foo = |baz;
```

I want to wrap `baz` in a call to `bar`. I type `bar(` and get:

```js
var foo = bar(|)baz;
```

I now call `sp-slurp-hybrid-sexp`:

```js
var foo = bar(|baz;)
```

whereas I wanted:

```js
var foo = bar(|baz);
```

A similar issues occurs in Python with `for i in len()foo.bar:` to `for i in len(foo.bar:)`.

Wilfred:
Fixes #534.

I think this is a reasonable change -- all of the tests still pass. I'm not sure what the `(newline)` call was intended to achieve. Removing it hasn't broken anything obvious to me.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -5919,7 +5919,6 @@ triggers that `sp-forward-slurp-sexp' does."
           (when (sp-point-in-blank-line)
             (delete-region (line-beginning-position) (1+ (line-end-position))))
           (sp-forward-sexp)
-          (sp-get (sp-get-hybrid-sexp) (goto-char :end-suf))
           (insert :cl :suffix))
         ;; TODO: move to hook
         (sp-get (sp--next-thing-selection -1)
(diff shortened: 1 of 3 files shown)


Fuco1:
I think the idea was in languages like C or Java to place the closing delimiter on an extra line?

A buffer

```c
foo() {}
int a = 1;
```

ends up as


```c
foo() {
  int a = 1;
}
```

Fuco1:
The tests against the closing delimiter are probably hacks to not do this in lisp.

```elisp
(foo)
foo (setq
     a b
     c d)
```

results in

```elisp
(foo
 foo (setq
      a b
      c d))
```

Wilfred:
Thanks. I'll open another PR once I've nailed down the corner cases.

PR #539, opened 2015-11-18
When slurping forward, don't include punctuation at end of line.

### LINKED ISSUE #534: sp-slurp-hybrid-sexp should not surround punctuation

Wilfred:
Fixes #534.

Let me know what you think of this approach. I've added tests, so I think the code is sane, but it's quite possible I've missed corner cases.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -5906,49 +5906,58 @@ Because the structure is much looser in these languages, this
 command currently does not support all the prefix argument
 triggers that `sp-forward-slurp-sexp' does."
   (interactive)
-  (-if-let* ((enc (sp-get-enclosing-sexp))
-             (bsexp (save-excursion
-                      (sp-get enc (goto-char :end))
-                      (when (sp-compare-sexps (sp-forward-sexp) enc >)
-                        (sp-get-hybrid-sexp)))))
-      (save-excursion
-        (sp-get enc
-          (goto-char :end-suf)
-          (delete-char (- (+ :cl-l :suffix-l)))
+  (let ((start-line (line-number-at-pos))
+        slurped-within-line)
+    (-if-let* ((enc (sp-get-enclosing-sexp))
+               (bsexp (save-excursion
+                        (sp-get enc (goto-char :end))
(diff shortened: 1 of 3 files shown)


Fuco1:
Sometimes you want to move the punctuation.  In C++ classes end with `};`, in many lanugages with lambdas you can have inline `;` or `,` which should move along (java, php, c#?).

I think this should be determined partially by syntax tables and partially by the user setting, similar to prefix.

Fuco1:
For a concrete example, consider

```c++
class foo {
public:
    int a;
};

int b = 7;
```

Fuco1:
It seems I don't understand the problem :D Because the example above works fine with your patch or without.

Wilfred:
Previously, we always included the punctuation at the end of the line. This patch introduces a new heuristic:

* If your open and close parens will be on the same line, don't include the punctuation. `foo(|)bar;` -> `foo(bar);`
* If your parens will be on different lines, do include the punctuation. `while () {|}\nfoo();` -> `while () {\n   foo();\n}`

This works really well in my testing. The only corner case I can think of is `bar(|)1.` -> `bar(1).`. I think this is a net improvement though.

@Fuco1 are you happy with this?

Fuco1:
I see.  Yes that sounds reasonable.

Could you please add the test case for C++ above? That tests that the punctuation *after* the closing delimiter moves along with it if necessary.  Then we can merge :)

Wilfred:
Test added :)

Fuco1:
Awesome, thanks a lot for your work! This tends to get annoying at times so I really appreciate someone digging into the crap :D

PR #545, opened 2015-11-29
Ensure that we can kill Rust words at the beginning of a buffer.

Wilfred:
Minor bug, added a test case.

@Fuco1 is this sufficiently boring that you're happy with me committing directly into master, or do you prefer PRs for fixes like this?


Fuco1:
Anything that goes into the "special" files where you know what you are doing you can push directly to master.  Especially since you're the author of the rust setup.

So long as you don't break tests I'm OK with pretty much any fix in the core as well, save for "I just rewrote 3000 lines of some legacy crap" :)

PR #550, opened 2015-12-12
Reworking README to showcase best features.

Wilfred:
Hi @Fuco1 

I didn't think that the existing README did smartparens justice. I've added some GIFs of some classic smartparens features, so new users can grok what smartparens offers them.

I've also reworded the README to make it easy to get up and running (something I found hard when I started using SP).

[Here's what it looks like](https://github.com/Fuco1/smartparens/blob/e7490481576cc0b53589ee35adc88e8e77b84660/README.md). What do you think?


README.md
```diff
@@ -5,30 +5,74 @@
...
+## Usage
 
-# Testing
+Inside Emacs, `M-x sp-cheat-sheet` will show you all the commands
+available, with examples.
 
-Smarparens uses [ert-runner](https://github.com/rejeep/ert-runner.el)
-for testing. To run all the tests, run:
+There's also a
+[great article that demonstrates movement and editing in smartparens](https://ebzzry.github.io/emacs-pairs.html)
```
Fuco1: Maybe change "that demonstrates" into "demonstrating", sounds a bit better to me.

Also add a period `.` to the end of this sentence (after wiki).

Fuco1:
Very nice! I've added one minor comment, but other than that let's go!

Fuco1:
Oh and you can also remove the gratipay badge while at it as that service basically shut down.

Wilfred:
Amended and merged! :)

PR #557, opened 2015-12-26
Interactive behaviour for sp-{next,previous}-sexp

### LINKED ISSUE #541: sp-next-sexp alternative fallback
"If there is no next expression at current level, jump one level up (effectively doing `sp-backward-up-sexp`)."

I'm not too keen on this fallback behaviour, what do you think about changing it to:

"If there is no next expression at the current level, jump one level up and retry." (which would have the effect of going to the next s-exp at whatever level has one)

fommil:
closes #541

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -4948,6 +4948,11 @@ Examples: (prefix arg in comment)
 
 (put 'sp-backward-sexp 'CUA 'move)
 
+(defcustom sp-interactive-dwim nil
+  "If non-nil and called `interactive', `sp-next-sexp' and
+`sp-previous-sexp' will automatically move the point to the
+end/beg of the current if point is not already on a beg or end.")
+
 (defun sp-next-sexp (&optional arg)
   "Move forward to the beginning of next balanced expression.
 
@@ -4956,6 +4961,10 @@ at current level, jump one level up (effectively doing
 `sp-backward-up-sexp').  Negative arg -N means move to the
 beginning of N-th previous balanced expression.
 
+If `sp-interactive-dwim' is non-nil, and this is called
+interactively, the point will be moved to the :beg before
+proceeding.
+
 With `sp-navigate-consider-symbols' symbols and strings are also
 considered balanced expressions.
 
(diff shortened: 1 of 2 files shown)


fommil:
green lights, so its got to be awesome!

I would add a test but I couldn't find one for next/previous

Wilfred:
@fommil I've pushed a few example tests, is that useful for you?

Looking at the code, could you please give the defcustom a more explicit name? DWIM can mean very different things to different people. Perhaps `sp-move-up-at-end`?

One other question: how is this different from `sp-forward-sexp`? I'm probably being stupid, but as far as I can see, `sp-forward-sexp` has this behaviour already.

fommil:
`sp-forward-sexp` places the point at the end of the sexp, doesn't it?

I'll add some tests and have a think about the defcustom name.

Wilfred:
Ah, I think Fuco's example mislead me.

```lisp
;; sp-next-sexp
(foo|) (bar) -> |(foo) (bar)

;; sp-forward-sexp
(foo|) (bar) -> (foo)| (bar)
```

Perhaps a more interesting case is:

```lisp
;; sp-forward-sexp
((foo|)) (bar) -> ((foo)|) (bar)

;; whereas you want to move straight to the next sexp:
((foo|)) (bar) -> ((foo)) |(bar)
```

I suppose you could write a `defcustom` like this?

```lisp
(defcustom sp-movement-at-end 'backward-up
  "Controls the behaviour when the user tries to move beyond
the last sexp at the current level with `sp-next-sexp'."
  :type '(radio
          (const :tag "Move backwards, up one level" 'backward-up)
          (const :tag "Move forwards to the next sexp, up many level if necessary" 'next-sexp)))
```

It might be nice if this took a movement function rather than symbols, but I don't think smartparens currently has a function for this exact behaviour you're looking for.

fommil:
umm, that's not what I'm trying to do. Have a look at my examples in the updated docs to show what I mean. I don't really care about the end/beginning behaviour.

fommil:
ok, almost done with tests. Are you ok with me making the defcustom on by default?

fommil:
ok, tests added. Any better ideas on the name? I can't think of anything.

Fuco1:
Wow, lots of work done here :) I'm really sorry but I'm totally wasted today, I will review this asap, but not now.  Thanks for taking it up Sam!

fommil:
aaah :christmas_tree: that time of year when its a race between passing out with all the drink, or tidying up all those `.emacs` niggles you've putting off for ages.

Fuco1:
The more I think about this I think the new behaviour should simply be the default.  We should also mark the function as "do not use from code" (as it's not a primitive like `sp-forward-sexp`).

I only used it once in `smartparens-html.el` and that can probably be easily fixed.

What about the case `((foo|)) (bar)` ? I think a useful invariant to introduce would be "point should never end back from where it started".  That way we will simply retry the procedure until point goes forward or there is nowhere to move (we're at the last form).

What do you think?

fommil:
Yup, that'd be nice! Now that Xmas is over I won't have so much time to do that though, sorry

Fuco1:
Most of the stuff you wrote seems to be reusable for this purpose, just throwing a couple loops around somewhere.

I'll try to merge it into a branch and work on that, then we merge from there.  If you want to add more changes it would be simpler to work in this repo directly, I'm OK with giving you push access.

fommil:
Cool. Can you please merge to master as is, just so I can use this?

fommil:
@Fuco1 any plans to merge?

fommil:
@fuco1 ping

Fuco1:
Right, I'll rewiew this over the weekend, thanks for the ping :P

fommil:
I only noticed because github added a new feature to show all your PRs in flight :smile:

fommil:
Heh spotted again thanks to more github upgrades. I'm surviving without this but it would be good.

Fuco1:
Looking at the test cases, it seems to me that this can be achieved by ignoring symbols:

    (setq-local sp-navigate-consider-symbols nil)

Try this in scratch buffer.  If not, can you add some case where it doesn't only "ignore" symbols?

fommil:
Might be a while, I'm on holiday and no access to a computer.

fommil:
I appear to have completely forgotten about this PR. I can't even remember creating it.

Fuco1:
Damn I suck.  This has been opened for two years now :/

I'm going to solve this right now :)

fommil:
awesome, although I thought I was the one holding it up so don't feel so bad :smile:

Fuco1:
Implemented in 40eccef.

The behaviour is such that with the option turned on the `sp-next-sexp` will keep going until the point ends up 

1) at the beginning of some sexp
2) being at a greater position than where it started from

The only exception is at the end of the buffer at the end of the sexp where there is no sexp forward, it will jump to the beginning of the last sexp in the buffer.

Analogically for the previous variant.

I renamed the option to `sp-navigate-interactive-always-progress-point` it is on by default and in the 2.0 release it will become non-configurable because it is simply the most logical behaviour.

THanks everyone for input!

fommil:
Awesome! Thank you so much :-)

Issue #481, opened 2015-06-16
help debugging when `sp--post-self-insert-hook-handler` stops working?


fommil:
After editing for a while (combination of scala and elisp), the insertion of closing brackets stops working and therefore `sp-local-pair` never triggers.

I'm pretty sure some major or minor mode that I'm using is disabling the character insertion hooks, but not during initial load because everything works fine for several minutes.

How can I start to debug what is happening when I type a bracket? Should I be looking any further than printing the current `post-self-insert-hook`?

Note to self, in `scala-mode`, when opening a fresh emacs and smartparens is working, this is the `post-self-insert-hook`

```
(scala-indent:fix-scaladoc-close scala-indent:indent-on-scaladoc-asterisk scala-indent:indent-on-special-words scala-indent:indent-on-parentheses sp--post-self-insert-hook-handler blink-paren-post-self-insert-function)
```

fommil:
renamed because I think this feels more like its failing at `post-self-insert-hook` and not getting to the `sp-local-pair`

Fuco1:
I'm not sure what you mean by `sp-local-pair` here, but yes, you need `sp--post-self-insert-hook-handler` in the hook for SP to work.  I would probably run a grep through my `~/.emacs.d/elpa` to see which packages mess with it.

fommil:
that's a good idea, I'll try that when it next fails and report back.

fommil:
and this is what `post-self-insert-hook` looks like when failing

```
(scala-indent:fix-scaladoc-close
 scala-indent:indent-on-scaladoc-asterisk
 scala-indent:indent-on-special-words
 scala-indent:indent-on-parentheses
 blink-paren-post-self-insert-function)
```

so `sp--post-self-insert-hook-handler` is being dropped.

I've `ag`d my entire `~/.emacs.d` for `post-self-insert-hook` and only `scala-mode2` is doing anything... and even then it looks pretty harmless https://github.com/hvesalai/scala-mode2/blob/master/scala-mode2-map.el#L14 (although I might turn off some of these now that I know about them).

I'm now stumped. I definitely have `Smartparens` (and in elisp `Smartparens-Strict`) mode enabled for the buffer.

Is there anything that wipes the `post-self-insert-hook`?

fommil:
I'm reopening this because it sounds like Emacs can remove the hook if it throws an error.

[...]

PR #558, opened 2015-12-26
Catch errors when running sp--post-self-insert-hook-handler

fommil:
This should help us to debug when this function is erroring, and will
ensure that Emacs does not remove the handler unexpectedly.

#481


Wilfred:
We've had fun with errors in hooks before, see #527. Can I suggest you use `with-demoted-errors`? This means `debug-on-error` still drops you into a debugger.

fommil:
if you want to debug an error in a `condition-case`, just turn on `debug-on-signal`. Did you experience something more than that?

fommil:
updated with `with-demoted-errors`

Wilfred:
Looks good to me, thank you! :)

PR #559, opened 2015-12-29
Pair < with > in parameterized Rust function calls.

meqif:
This fixes an oversight in pairing angle brackets in Rust when parameterizing function calls, such as `iterator.collect::<Vec<u32>>()`.

By the way, thank you for this excellent package. :smile:


Wilfred:
Thank you :)

Issue #236, opened 2013-11-21
sp-forward-slurp-sexp keeps space after itself in python-mode


k-bx:
Imagine yourself having this python code:

```
foo(bar)
```

You want to place `boo()` on top of this call. What I do is (`|` is a cursor):

```
|foo(bar)
boo|foo(bar)
boo(|)foo(bar)
```

Then I press `C-rightarrow` (`sp-forward-slurp-sexp`) and it transforms into:

```
boo(|foo)(bar)
```

which is ok. But then I again press `C-rightarrow` and get:

```
boo(|foo (bar))
```

Which leaves a space after foo.

Fuco1:
Related to #404.  We should come up with a solution there.

c-oreills:
This is not actually a duplicate - #404 has been fixed (relating to . characters moving) but the testcase at the top of this file still fails.

The code above could be included in smartparens-python

Fuco1:
Thanks for catching this.  By code above you mean the code I wrote or something else?  Have you verified it works?

Wilfred:
FWIW, `sp-slurp-hybrid-sexp` doesn't suffer from this problem.

c-oreills:
Thank you, that's solved it! =)

Fuco1:
And we only had to wait 2 years :D

Well, thanks everyone for chipping in!

PR #561, opened 2016-01-08
Don't insert whitespace when slurping in Python.

Wilfred:
Hi @Fuco1

This is just applying the fix you suggested for #236. I think just doing `(` and `[` is reasonable here. I think the code is correct. It's using dynamic scoping so generates a warning, but it does pass the tests.

I've also committed 62429854f7ea919b9610bacb282e756f0bc5befc to make hybrid slurping more discoverable.


Fuco1:
Yea, the dynamic scoping is unfortunate... I'm not quite sure if it is possible to retroactively change the api now though... the context variable for example is pretty much never used as was intended.

PR #565, opened 2016-01-28
Fix python-indent-dedent-line-backspace in strict mode

### LINKED ISSUE #552: Incorrect paren deleted in python-mode with smartparens-strict-mode
```python
return foo(bar, lambda x: baz(x))|
```

Pressing backspace will remove the end paren, leaving the parens unbalanced.

PythonNut:
This should fix #552.

### DIFF
--- a/smartparens-python.el
+++ b/smartparens-python.el
@@ -72,5 +72,15 @@
         (goto-char (sp-get ok :end))
         (when (looking-back " ")
           (delete-char -1))))))
+
+(defadvice python-indent-dedent-line-backspace
+    (around sp-backward-delete-char-advice activate)
+  (if smartparens-strict-mode
+      (cl-letf (((symbol-function 'delete-backward-char)
+                 (lambda (arg &optional killp)
+                   (sp-backward-delete-char arg))))
+        ad-do-it)
+    ad-do-it))
+
 (provide 'smartparens-python)
 ;;; smartparens-python.el ends here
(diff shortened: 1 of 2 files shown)


Wilfred:
Works beautifully! :) Would you mind adding a test?

Fuco1:
You don't test whether the strict mode is enabled.

PythonNut:
@Fuco1 ah, I thought that `sp-backward-delete-char` did the testing.

Fuco1:
Me too, but it seems it doesn't.  Maybe we should add it there as a separate patch?

Fuco1:
Hi.  I would prefer you add the test to the advice, because I can imagine sometimes you might want to call the method outside of the strict mode (from code).  Plus it only activates in strictmode anyway (via the keymap) so having the test there would be a duplication.

In case of misunderstanding, I think @Wilfred ment a test -> a unit test, I ment a test for the setting in code (that is an `if` condition or whatever).

If you don't know how to add a regression test that's my fault as I still haven't written the documentation :/

PythonNut:
> In case of misunderstanding, I think @Wilfred ment a test -> a unit test, I ment a test for the setting in code (that is an if condition or whatever).

Ah. I see.

> If you don't know how to add a regression test that's my fault as I still haven't written the documentation :/

Yeah, I'll look into it. I'm not that familiar with testing in Elisp, but I think I looked at the code the other day and it wasn't too bad.

PythonNut:
@Wilfred I've added a unit test.

@Fuco1 I've added a runtime check for `smartparens-strict-mode`.

Fuco1:
Great, thanks!

Wilfred:
Thanks @PythonNut :)

PR #568, opened 2016-02-06
Racket: Don't pair quote

greghendershott:
I think quote should be disabled, just like quasiquote already is.

See also: https://github.com/greghendershott/racket-mode/issues/200

### DIFF
--- a/smartparens-racket.el
+++ b/smartparens-racket.el
@@ -48,6 +48,7 @@
 (require 'smartparens)
 
 (sp-with-modes '(racket-mode racket-repl-mode)
+  (sp-local-pair "'" nil :actions nil)
   (sp-local-pair "`" nil :actions nil)
   (sp-local-pair "#|" "|#"))
 


Wilfred:
Fair enough :). Would you mind adding a test?

greghendershott:
> Would you mind adding a test?

Honestly? I would mind. Why:

0. I don't want to install Cask only for this tiny PR.
1. I'm not sure what's a meaningful test:
    - This only changes a default, and only changes it do nothing when <kbd>'</kbd> is typed.
    - racket-mode doesn't do anything special with <kbd>'</kbd>, it's just `self-insert-command`.
    - I can't find an example of a test for the existing default of leaving backtick (a.k.a. quasiquote) alone: `` (sp-local-pair "`" nil :actions nil) ``. Maybe I overlooked it. Or maybe there isn't such a test because of the previous two points? :)

This is one of those PRs where the spirit is, "I could opened an Issue, but, a PR shows what I mean, and if we're all lucky, you can just merge it. Otherwise please consider it an issue, and use your better understanding to do it some better/preferred way." :)

Wilfred:
Tests are great as they give an example of the expected behaviour, and ensure we don't break it in future. A test would look like this:

```lisp
(require 'racket-mode)
(require 'smartparens)

(ert-deftest sp-test-racket-quote ()
  "' is a not a paired delimiter in Scheme."
  (sp-test-with-temp-buffer "(print |)"
      (racket-mode)
    (execute-kbd-macro "'x")
    (should (equal (buffer-string) "(print 'x)"))))
```

However, this test already passes! smartparens-config.el already contains the following:

```lisp
(sp-with-modes sp-lisp-modes
  ;; disable ', it's the quote character!
  (sp-local-pair "'" nil :actions nil)
```

so I'm not sure what's going on.

Fuco1:
Seems to me the person in the original issue has some rather old version.  The issue with the delay they report seems very weird, that should not affect the behaviour of the parser at all.

Wilfred:
I can reproduce this an empty .emacs.d, installing racket-mode and smartparens. In my normal config I can't reproduce it though.

Wilfred:
OK, I'd forgotten to do `(require 'smartparens-config)`. I don't think there's a bug here.

Wilfred:
Since this works already, I'm going to close this. Please make sure you load the default configuration with `(require 'smartparens)`, as shown in https://github.com/Fuco1/smartparens#getting-started

Let me know if we've missed anything.

greghendershott:
I see. Yes. Looks like our mutual user and I both misunderstood the smartparens config. Sorry about the confusion (and also sorry to be slow to follow up on this).

IIUC now, a good config would simply be: `(require 'smartparens-config)`.  What the FM already says. :)

Because already `smartparens-config`:

- disables pairing `'` for lisp modes (because `sp-lisp-modes` includes `'racket-mode` by default; thanks!)
- does `(eval-after-load "racket-mode" '(require 'smartparens-racket))`, for extra Racket specifics.

So we're all set.

I had updated the racket-mode README to suggest doing `(require 'smartparens-racket)`. Wrong. I'll change that to say "follow the smartparens instructions, including `(require 'smartparens-config)`".

PR #574, opened 2016-02-16
Ensure that we pair ' in code in Python.

### LINKED ISSUE #573: Change to smartparens-python.el breaks single quote pairing
Hi,

This is in reference to: 065fa38e9e2c10a26b523016d521e0be0a021e0b.

The prior (and desired) behavior was that a single quote would generate an auto-pair when in python mode. With this change, no pair is generated when I hit the single quote key in python mode. I have confirmed that this latest change caused the change in behavior, i.e, when I remove the change the behavior returns to normal.

To me this is a regression. Could the change please be undone or corrected?

Wilfred:
Fixes #573 and adds a test.

@Fuco1 I'm not sure if this is the right approach. The naive implementation we had before didn't work, but I'm not sure if this is a bug in the smartparens parsing.

```python
# we start with this
x = |

# the user presses '
x = '|

# however, smartparens thinks that point is now in a string, 
# so didn't insert the matching ' !
```

Is there a better approach? Moving around and calling `syntax-ppss` seems hacky but I can't see a better way using just `id` `context` and `action`.

### DIFF
--- a/smartparens-python.el
+++ b/smartparens-python.el
@@ -53,9 +53,17 @@
 (--each '(python-mode inferior-python-mode)
   (add-to-list 'sp-sexp-suffix (list it 'regexp "")))
 
+(defun sp-python-in-string-p (id action context)
+  "Return non-nil if point is in a string, taking
+care not to be confused by an unclosed ' that's just been typed."
+  (and (eq context 'string)
+       (save-excursion
+         (backward-char 1)
+         (nth 3 (syntax-ppss)))))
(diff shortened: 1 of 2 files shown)


Fuco1:
The problem is that python thinks of both `"` and `'` as strings.  The regular `:unless` test would work otherwise.

I think your solution is good for the meantime.  What we could do is to send some additional information for the context, such as what delimiter actually starts the string.

Issue #572, opened 2016-02-16
sp-backward-kill-sexp should remove punctuation too


Wilfred:
I'm trying to kill expressions in non-sexp languages (this example is from Python):

```
# Starting with this:
foo = bar.baz - biz()|

# sp-backward-kill-word ignores delimiters
foo = bar.baz - biz()|
foo = bar.baz - |()

# sp-backward-kill-symbol ignores delimiters too
foo = bar.baz - biz()|
foo = bar.baz - |()

# sp-backward-kill-sexp does what I want for delimeters, but doesn't like infix -:
foo = bar.baz - biz()|
foo = bar.baz - biz|
foo = bar.baz - |
foo = bar. - |
```

Previously, I used `backward-kill-word`. However, that falls down with functions with arguments:

```
# backward-kill-word produces unbalanced parens
foo = bar.baz - biz(quux, qox)|
foo = bar.baz - biz(quux, |
```

I think I should be using `sp-backward-kill-sexp`. Is it intended to work for non-lisps?

Fuco1:
I'm not sure what you're after :O Do you want to kill the entire `biz()` thing in one go?

Wilfred:
Yep, I want to kill the whole `biz()` in one go. However, `sp-backward-kill-sexp` is confused by infix operators:

```
foo = bar.baz|
foo = bar.|
foo = .|

x - y|
x - |
 - |
```

`sp-backward-kill-sexp` is nearly perfect, but it's confused in these cases. Is this command intended to work in non-lisps (so this would be a bugfix/improvement) or do we need a `sp-backward-kill-hybrid-sexp`?

Fuco1:
I'm not sure, but I think nuking punctuation in the way isn't a terrible idea.

PR #579, opened 2016-02-20
 Bugfixes and improvements to `sp-kill-sexp`

Wilfred:
This pull request includes:

* Tests for `sp-kill-sexp` based on the examples in the docstring
* A fix for a bug I found when testing those examples (SP is confused about negative prefix arguments)
* A fix for #572 (handling punctuation) by starting the kill from the current point position

I've not explored this part of the code before, so please let me know if I've done anything silly :)

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -5511,7 +5511,7 @@ Note: prefix argument is shown after the example in
          (n (abs arg))
          (ok t)
          (b (point-max))
-         (e (point-min))
+         (e (point))
          (kill-fn (if dont-kill 'copy-region-as-kill 'kill-region)))
     (cond
      ;; kill to the end or beginning of list
--- a/test/smartparens-commands-test.el
+++ b/test/smartparens-commands-test.el
@@ -3,7 +3,8 @@
    ((not (boundp 'mode)) (emacs-lisp-mode))
    ((eq mode 'elisp) (emacs-lisp-mode))
    ((eq mode 'racket) (racket-mode))
-   ((eq mode 'c) (c-mode)))
+   ((eq mode 'c) (c-mode))
+   ((eq mode 'python) (python-mode)))
   (smartparens-mode 1))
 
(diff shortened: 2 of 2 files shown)


Fuco1:
I test commands using the `sp-test-command` macro, it's in the `smartparents-commands-test.el` file.  It has the added bonus of automatically checking if point ends up where we want, and you can perform sequences of multiple steps along the way.

Fuco1:
And which example did not work properly?  I'm not sure testing for 4 or negative makes sense in the way I intented the command to work.

Wilfred:
4675763b0fffe4e46c9cfbe97b907c419ec910cc shows the example that doesn't work. The relevant lline from the docstring of `sp-kill-sexp`:

```
 (1 2 3 4| 5 6)    -> (|5 6)      ;; - \\[universal-argument]
```

Fuco1:
Works for me, you need to do: `C-- C-u sp-kill-sexp`

Wilfred:
Yikes, I had no idea you could pass negative prefix arguments like that. I was doing `C-u - M-x sp-kill-sexp`. Should that do the same thing? That was what I was expecting.

Fuco1:
Nope, not the same thing.  `C-u` initiates the "prefix argument map" in which numbers specify the number (so you just do `C-u 2` instead of `C-u C-2`... or just `C-2`) and - specifies negation *in that mode*.  If you do `C--` before, it stores the negative and then `C-u` sends in `(4)` which gets negated---or if you follow up with numbers it will pass onto the numbers.

It's a bit crazy :P

Wilfred:
OK, that kind of makes sense :). I'll update the PR.

Wilfred:
Updated, let me know what you think.

Fuco1:
Seems to work so I'm OK with merging this.

You rewriting test from the docstrings gave me the idea that maybe we could auto-generate tests from the documentation (kind of like dash)... but I'm not sure if it's worth the trouble. (haskell also has something similar)

PR #582, opened 2016-02-25
sp-select-next-thing should handle buffer limits gracefully.

### LINKED ISSUE #449: `wrong-type-argument number-or-marker-p nil` in sp-select-next-thing when no thing to select
Placing cursor before end of empty buffer (e.g. minibuffer) and using sp-wrap-with-pair generates an error in `sp-select-next-thing`. It should fail gracefully without an error.

Full backtrace:
```
Debugger entered--Lisp error: (wrong-type-argument number-or-marker-p nil)
  sp-select-next-thing(1 nil)
  sp-select-next-thing-exchange(1 nil)
  sp-wrap-with-pair("(")
  (lambda (&optional arg) (interactive "P") (sp-wrap-with-pair "("))(nil)
  call-interactively((lambda (&optional arg) (interactive "P") (sp-wrap-with-pair "(")) nil nil)
  command-execute((lambda (&optional arg) (interactive "P") (sp-wrap-with-pair "(")))
```

Wilfred:
Fixes #449.

### DIFF
--- a/test/smartparens-commands-test.el
+++ b/test/smartparens-commands-test.el
@@ -1,3 +1,5 @@
+(require 'smartparens)
+
 (defun sp-test-command-setup ()
   (cond
    ((not (boundp 'mode)) (emacs-lisp-mode))
@@ -538,5 +540,13 @@ be."
       (sp-test-with-temp-elisp-buffer "(fo|o)"
         (let ((unread-command-events (list ?\a)))
           (call-interactively 'sp-rewrap-sexp))
+        (error "We should never get here"))
+    (user-error t)))
+
+(ert-deftest sp-test-command-sp-select-next-thing-empty-buffer ()
+  "Ensure we call user-error at buffer end."
+  (condition-case c
+      (sp-test-with-temp-elisp-buffer ""
+        (call-interactively 'sp-select-next-thing)
         (error "We should never get here"))
     (user-error t)))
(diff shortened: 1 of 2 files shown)


test/smartparens-commands-test.el
```diff
@@ -540,3 +542,11 @@ be."
           (call-interactively 'sp-rewrap-sexp))
         (error "We should never get here"))
     (user-error t)))
+
+(ert-deftest sp-test-command-sp-select-next-thing ()
```
Fuco1: Won't this clash with the generated test and override it?

Wilfred: Do we have a test for this command? I couldn't see one.

Fuco1: The `sp-test-command` macro generates ert tests with names `sp-test-command-<name>`, so it would clash there.  Just add something at the end, like `-empty-buffer`.  For example see `sp-test-command-sp-rewrap-sexp-invalid-pair` in `...-commands-test.el`

Wilfred: OK, so you're saying we don't have a test, but if we use `sp-test-command` in future it would clash with this test.

Gotcha, test name updated.

Fuco1:
Hm, you could've merged this in yourself :D Anyway, I'm doing some issue/pr cleanup now, so this is good to go, thanks!

PR #583, opened 2016-02-25
Add code coverage

Wilfred:
It would be nice to track coverage with our tests, so I've set up code coverage. Let me know what you think :)

You can see an example of the coverage data here: https://coveralls.io/github/Wilfred/smartparens


Fuco1:
Sweet, I wanted this for some time but had no idea how to go about it.

PR #600, opened 2016-05-07
Add smartparens support for motion-mode.

jojojames:
motion-mode (used primarily for rubymotion) is a derived major mode of ruby-mode.


Fuco1:
Thanks!

PR #605, opened 2016-05-16
Add memoization for sp--strict-regexp-quote

### LINKED ISSUE #603: Memoize sp--strict-regexp-quote
One potential improvement to #595 would be to memoize `sp--strict-regexp-quote`. According a reverse rendered profile it's on the higher end of cpu samples. It's a pure function with a limited input set, so it's ripe for memorization. I'd be happy to offer a pull, but I don't know how you'd like to memoize it, if you'd like to pull in a package like https://github.com/skeeto/emacs-memoize or just roll your own. 

```
+ sp--looking-back                                                269  16%
+ sp--strict-regexp-quote                                         156   9%
+ syntax-ppss                                                     151   9%
+ sp--looking-back-p                                              111   6%
  flyspell-post-command-hook                                       76   4%
+ get-frame-persp                                                  71   4%
+ sp-delete-pair                                                   55   3%
+ spaceline-ml-main                                                53   3%
+ sp--looking-at                                                   48   2%
+ sit-for                                                          39   2%
+ #<compiled 0x40ebed8b>                                           38   2%
+ intern                                                           30   1%
+ timer--time-less-p                                               30   1%
+ pl/create-or-get-cache                                           24   1%
```

aaronjensen:
Fixes #603

To be honest, this didn't seem to make much difference. I couldn't reproduce `sp--strict-regexp-quote` taking up a large number of samples while profiling and adding the memoization in didn't seem to help. Furthermore, the timers created by the memoize package seem to dominate the profiler without actually seeming to impact typing latency (which makes sense, I guess). If I disabled the timers by setting `(setq memoize-default-timeout nil)` they didn't appear in the profile, but since that's a global setting I don't think it makes sense for us to set. It doesn't seem like there's a way with this package to say you don't a timeout for a specific function because of the `(or timeout memoize-default-timeout)`.

Long story short, it'd be great if someone else could measure with [typometer](https://github.com/pavelfatin/typometer) the difference between memoizing and not. It seemed to not matter too much on my machine.

Lastly, if we were going to memoize, I don't think this is the package for it because we probably don't actually want a timeout for this specific function. Adding so many additional timers probably doesn't help anything.

### DIFF
--- a/Cask
+++ b/Cask
@@ -6,6 +6,7 @@
 
 (depends-on "cl-lib" "0.3")
 (depends-on "dash" "2.12.1")
+(depends-on "memoize" "1.0.1")
 
 (development
   (depends-on "f")
--- a/smartparens-pkg.el
+++ b/smartparens-pkg.el
@@ -1,3 +1,4 @@
 (define-package "smartparens" "1.7.1" "Automatic insertion, wrapping and paredit-like navigation with user defined pairs."
-  '((dash "2.12.1")
+  '((memoize "1.0.1")
+    (dash "2.12.1")
     (cl-lib "0.3")))
--- a/smartparens.el
+++ b/smartparens.el
@@ -53,6 +53,7 @@
 (require 'cl-lib)
 (require 'dash)
 (require 'thingatpt)
+(require 'memoize)
 
 (eval-when-compile (defvar cua--region-keymap))
 (declare-function cua-replace-region "cua-base")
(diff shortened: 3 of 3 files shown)


Fuco1:
Hm, so if the effect is minimal, I will postpone merging this just yet.  No need to bring in an extra complexity if it is for nothing.  Thanks for testing it though, I'll leave this open in case we decide to go through with it.

PR #612, opened 2016-05-29
Removed pairing for backtick and apostrophe.

### LINKED ISSUE #611: LaTeX mode single quotes
There seems to be a confusing with single quotes in LaTeX. When a single quote is entered, the minibuffer shows "Search failed. This means there is unmatched expression somewhere or we are at the beginning/end of file."

Moreover, I still see that the backtick  is still matched with ' even though the line `  (sp-local-pair "'" nil :actions nil)` exists in `smartparens-latex.el`.

I tried removing the pair via `(sp-pair "'" nil :actions :rem)`, but that has no apparent effect.

legendre6891:
This fixes #611.

It seems that sp-latex-skip-match-apostrophe doesn't perform its stated
function, because placing the cursor before a quote (say, in math mode,
where quote is used to denote a derivative) will result in "Unmatched
expression." showing up in the minibuffer.

Note that smartparens, in an attempt to find the matching backtick for
the single quote, substantially slows down editing in any (La)TeX file
containing unmatched quotes (e.g. derivative f'(x) showing up in math mode).

Since single quotes appear relatively infrequently in English prose, the
loss in functionality is minimal and the speed-up benefit seems to be
worth it.

### DIFF
--- a/smartparens-latex.el
+++ b/smartparens-latex.el
@@ -67,12 +67,6 @@
       (goto-char (sp-get sp-last-wrapped-region :beg-in))
       (insert " "))))
 
-(defun sp-latex-skip-match-apostrophe (ms mb me)
-  (when (equal ms "'")
-    (save-excursion
-      (goto-char me)
-      (looking-at-p "\\sw"))))
-
 (defun sp-latex-skip-double-quote (_1 action _2)
   (when (eq action 'insert)
     (when (looking-at-p "''''")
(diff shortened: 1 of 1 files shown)


Fuco1:
Hi

If the pair doesn't work in prose, then we should fix the quote in math mode.  In prose the backtick pair functions normally and is super helpful.  You can do conditional unless/when on various actions like navigation, search, highlighting, nisertion etc.

Fuco1:
I fixed this by adding an unless predicate

Issue #593, opened 2016-04-17
sp-forward-slurp-sexp should know about Clojure set literal, lambda


lvh:
Example, `|` being point

```
(abc|) #{1 2 3}
```

`sp-forward-slurp-sexp` results in:

```
(abc| #){1 2 3}
```

Doing it a second time results in:

```
(abc| # {1 2 3})
```

(having introduced a space between `#` and `{`).

Instead, I would have liked it to:

```
(abc| #{1 2 3})
```

The same thing is true for Clojure lambda expressions, e.g. `#(inc %)`; where the `#` is captured separately.

I think smartparens currently gives me the configurability to fix this for my setup, but I wanted to report it upstream since this seems generally useful.

Fuco1:
It should be enough to do what we did for racket, that is adding the # as a prefix symbol.

Thanks for reporting, it's a nice feature to add.

lvh:
This also affects conditional read statements for `cljc` files, which have `#?(...)` and `#?@(...)` which should be interpreted as one token.

Fuco1:
Where could I find all the possible prefixes clojure supports?

lvh:
That's a good question! Here's an attempt at an exhaustive list:

Reader macros (http://www.clojure.org/reference/reader):

- `#{...}` (sets)
-  ^{:a 1}` (metadata maps , appear to be handled correctly already)
- `#"..."` (regex pattern)
- `#'xyzzy` (varquote)
- `#(xyzzy 1 2 %)` (anonymous fn)
- `#_` (comment reader macro). This comments the next form. This appears to be an edge case, since it correctly captures the `#_` entirely, but it doesn't capture the thing it comments out... Perhaps that's OK; the user can just slurp twice :) Same thing with tagged literals.
-  Reader conditionals, e.g.:

```clojure
#?(:clj     Double/NaN
   :cljs    js/NaN
   :default nil)
```

and 

```clojure
#?@(:clj [3 4] :cljs [5 6])
```

Fuco1:
The issue is that some of those characters are correctly set as `prefix` class by the clojure mode but others are not.  Since characters can only have one class, in SP we do the ugly thing and just recognize prefixes by a regexp... that's why I asked :)  I would also need all the valid combinations (for example in elisp you can have a prefix `',@` and theoretically it can be infinitely nested (like `'''''',@'''`).

[...]

PR #620, opened 2016-06-23
[#593] Add smartparens-clojure.el

vspinu:
Make 'sp-sexp-prefix' aware of Clojure constructs.

### DIFF
--- a/smartparens-clojure.el
+++ b/smartparens-clojure.el
@@ -1,0 +1,49 @@
+;;; smartparens-clojure.el --- Additional configuration for Clojure mode.
+;;
+;; Author: Vitalie Spinu <spinuvit@gmail.com>
+;; Maintainer: Matus Goljer <matus.goljer@gmail.com>
+;; Created: 14 July 2016
+;; Keywords: abbrev convenience editing
+;; URL: https://github.com/Fuco1/smartparens
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License:
+;;
+;; This file is part of Smartparens.
+;;
+;; Smartparens is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+;;
+;; Smartparens is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with Smartparens.  If not, see <http://www.gnu.org/licenses/>.
+;;
+;;; Commentary:
+;;
+;; This file provides some additional configuration for Clojure mode.  To use
+;; it, simply add:
+;;
+;; (require 'smartparens-clojure)
+;;
+;; into your configuration.  You can use this in conjunction with the
+;; default config or your own configuration.
+;;
+;;; Code:
+
+(require 'smartparens)
+
+(defvar sp-clojure-prefix "\\(?:[@`'#~,_?^]+\\)"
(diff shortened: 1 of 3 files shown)


vspinu:
To be clear about the issues. The `(sp-wrap-with-pair "(")` works as expected on the following:

```clojure
#{...}
^{:a 1}
#(xyzzy 1 2 %)
#_(comment reader macro)
(#?(:clj     Double/NaN
    :cljs    js/NaN
    :default nil))
~unquote-me
`~'unquote-me
~@(splice-me)
`(quote-me)
```quote-me
@(deref-me)
@deref-me
#?@(:clj [3 4] :cljs [5 6])
```

but fails on 

```clojure
#"..."
#'xyzzy
```

`sp-forward-barf-sexp` has issues with all prefixes that include `#`. For example `(|#{...})` and `sp-forward-barf-sexp` creates `(#){...}` instead of `()#{...}`. Still better than nothing.

smartparens-clojure.el
```diff
@@ -0,0 +1,45 @@
+;;; smartparens-clojure.el --- Additional configuration for Clojure mode.
+;;
+;; Author: Matus Goljer <matus.goljer@gmail.com>
```
Fuco1: You are the author, not me.

smartparens-clojure.el
```diff
@@ -0,0 +1,45 @@
...
+;; Maintainer: Matus Goljer <matus.goljer@gmail.com>
+;; Created: 8 February 2015
```
Fuco1: Date

Fuco1:
It would be most awesomest if you could add an ert test for the examples you gave above.

Here's a template you can use

```elisp
(ert-deftest sp-test-somethingsomething-clojure ()
  (sp-test-with-temp-buffer "initial buffer content, | is the initial point"
      (clojure-mode)
    (your-code-in-a-new-temp-clojure-buffer)
    (optionally (insert "|") so you can test the resulting point)
    (should (equal (buffer-string) "output | buffer"))))
```

See the python tests for more inspiration.

Fuco1:
@vspinu You should probably open a separate issue about the barf behaviour, I smell a general problem there.  Thanks!

smartparens-clojure.el
vspinu: > You are the author, not me.

This is a trivial copy paste change of your code and suggestions. I am a bit wary of 'Maintainer' field as I am a somewhat intermitent clojure user and might not hold to the promise. 

In any case I can un-modestly put my name there if you prefer so.

Fuco1: You can leave me as the maintainer :P

Fuco1: It's just for the legal nonsense in the case we eventually get into emacs proper... other authors will be added as they contribute to the file.

vspinu:
>  could add an ert test 

Sure. I will add ert tests. 

I see some corner cases with the current regexp. I would like to dig a bit more into smartparens inner workings first.

Fuco1:
Hey, just checking in, any progress? :)

vspinu:
Unfortunately not.  Sorry about this. I will try to get this rolling by the end of the week.

Fuco1:
No problem :)

vspinu:
I think I am done  for now. A bunch of tests are commented out as they break because of #621. 

Unfortunately by simply setting `#` to prefix syntax doesn't solve all issues. If the prefix ends with a  symbol characters  (like in `#?(....)`)  the operations will still fail. 

 I don't know what would be the proper fix for this. An alternative to `sp-sexp-regexp` might be to introduce a special syntax table that would be active during sp commands.

Fuco1:
Hm, that is odd, I think the regexp shouldn't care about syntax properties.

Anyhow, let's merge this and I'll add you as a contributor so you can take care of the clojure stuff if you find some more issues.  You can also try to work on 621 but that seems more of an internal issue, so don't feel pressured :)

Thanks for the contribution.

PR #631, opened 2016-07-15
Prevent adding new whitespace when slurping forward/backward in OCaml and Latex

taquangtrung:
Hi,

When using the excellent slurping feature, I encountered a problem in both OCaml and Latex. That is, a new whitespace is added when it is not supposed to:

For example: given a string `(foo) List.map`, I want to slurp forward to obtain `(foo List.map)`,
the current slurping forward adds a new space between `List.` and `map` as follow:

```
(foo) List.map  ~~> (foo List).map ~~> (foo List .map)
```

The unwanted space should be removed, so that the expected behaviour is:

```
(foo) List.map  ~~> (foo List).map ~~> (foo List.map)
```

In these commits, I fixed the this problem for both forward and backward slurping, in major modes for OCaml and Latex.

Please fell free to merge them into your repository, if they are necessary.

Bests,
Trung.

### DIFF
--- a/smartparens-config.el
+++ b/smartparens-config.el
@@ -116,6 +116,7 @@
   (eval-after-load it '(require 'smartparens-python)))
 (eval-after-load "scala-mode" '(require 'smartparens-scala))
 (eval-after-load "racket-mode" '(require 'smartparens-racket))
+(eval-after-load "tuareg-mode" '(require 'smartparens-ocaml))
 
 (provide 'smartparens-config)
 
--- a/smartparens-latex.el
+++ b/smartparens-latex.el
@@ -90,10 +90,32 @@ This predicate is only tested on \"insert\" action."
 (add-to-list 'sp-navigate-skip-match
              '((tex-mode plain-tex-mode latex-mode) . sp--backslash-skip-match))
 
+
+(defun sp-latex-pre-slurp-handler (id action context)
+  ;; If there was no space before or after, we shouldn't add on.
+  ;; Variable ok, next-thing are defined in
+  ;; `sp-forward-slurp-sexp' and `sp-backward-slurp-sexp'
+  (when (eq action 'slurp-forward)
+    (save-excursion
+      (when (and (sp-get ok (/= :len-in 0))
+                 (= (sp-get ok :end-suf) (sp-get next-thing :beg-prf)))
+        (goto-char (sp-get ok :end))
+        (when (looking-back " ")
+          (delete-char -1)))))
+  
+  (when (eq action 'slurp-backward)
+    (save-excursion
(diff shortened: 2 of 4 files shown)


Fuco1:
Wow, this looks amazing on the first sight!

Could I bother you to add test cases as well?  You can inspire from the python mode test file.

I'll try to review this ASAP.

taquangtrung:
Sure! I just added a simple test cases for OCaml mode.
I don't know how to manually test it, so I'm looking forward to your automatic testing result first.

Btw, I found out that the problem of adding more unwanted space could happen in many major modes, so fixing it by patches for these mode are quite redundant (in my opinion).

I believe the root cause for `forward-slurping` is because of the following two lines in `sp-forward-slurping-sexp`

    (insert " ")
    (setq ins-space -1)

If the first line is removed, and the second line is changed to `(setq ins-space 0)`, then the extra whitespace will not be added.

How do you think about it? 
I don't know if this modification may affect some major modes.

Fuco1:
The space-adding code is there because of lisps, because something like `(foo)bar` has to end up as `(foo bar)` rather than `(foobar)`.

Maybe if we make the number adjustable and change it in different modes, that might be helpful too.

taquangtrung:
Thanks for the explanation! You are right about `lisps`. I have this idea about checking the sexp before and after the parenthesis or the square bracket:

Given that `A` and `B` are two sexp and we want to slurp the parenthesis in `A)B`:
  * If either `A` or `B` are delimiter, i.e, one of the `. ; , + - *`, etc, then do not add new whitespace
  * If both `A` and `B` are words (contains only letters or digits), then add new whitespace in the middle of them.

How do you think about it?

PythonNut:
So I'm dumb, and now we have https://github.com/PythonNut/smartparens/commit/cf367628e04e8e2336cf900302a874eabac809e3, which is basically just the LaTeX portion of this, but applied to more pairs.

Fuco1:
@PythonNut do you want to provide it as a separate patch for now? I'll probably review this behaviour and try to come up with something universal for 1.10 release, but I would accept a temporary fix for now.

Fuco1:
@taquangtrung I rebased this branch on master and the tests are failing, can you check it out? https://travis-ci.org/Fuco1/smartparens/jobs/173359035

I would like to merge this before 1.9 is out :)

PythonNut:
> @PythonNut do you want to provide it as a separate patch for now? I'll probably review this behaviour and try to come up with something universal for 1.10 release, but I would accept a temporary fix for now.

Sure. Let me make a PR?

Fuco1:
I rebased and merged this, the issue was simple for me to fix on my own. Thanks for your contribution!

PR #640, opened 2016-08-21
Add local pair for norm bars

cqql:
The title says it all. I had to write lots of norms lately so I added another local pair.


Fuco1:
Thanks!

PR #648, opened 2016-09-12
Don't autocomplete the lua blocks when inside comments

bunelr:
Hello, 

First of all, thanks for smartparens, I use it all the time :)

I have some small gripes with smartparens-lua:

At the moment, if you're in a Lua buffer and you are writing comments, 
```lua
-- Do this for|
```
 and you press `SPC`, you'll get autocompleted

```lua    
-- Do this for | do
    end
```

which is problematic because you have to watch out for the words you use in the comments to avoid getting spurious completion.

I don't see a case where you would want to have this auto-completed when you're writing comments but I might be wrong.

What do you think?


Fuco1:
I think this is reasonable :)

Do you know how to write tests?  You can look into `tests/...python-test`, but don't forget to add `(require 'smartparens-lua)` on top :)  If not that's fine too.

bunelr:
Sure, sorry I didn't include any. 
I'll add some tests later tonight when I come home.

(I don't think the current failing test is related though)

Cheers

Fuco1:
The current test fail is some magical incompatibility with emacs 24.3.  I'll need to figure out how to label it "allowed to fail" :o  Thanks.

bunelr:
Tests added :)

Fuco1:
Great! Thank you very much.

bunelr:
Resolved the merge conflict by merging Master back into this branch.
The problem was in the Cask file, addition of required elements were added in 9eed75819e15d6a088bb41cbbc8755deff833d4f at the same time.

Fuco1:
Great, thanks for the contribution!

PR #651, opened 2016-09-23
Fix autoskipping closing chevrons in rust-mode

### LINKED ISSUE #642: Closing angle bracket in rust-mode not autoskipping
`emacs -Q`, with `(require 'smartparens-config)`, then `rust-mode` and `smartparens-mode`.

Type `T<`, and smartparens adds a matching closing angle bracket: `T<|>`.  Type `E>` and you get `T<E>|>`, regardless of the setting of `sp-autoskip-closing-pair`.

Same thing happens without the `E`.

smartparens version:  20160821.152
rust-mode version: 20160820.255
Emacs version: 24.5.1

fmdkdd:
Before this commit, closing chevrons were not autoskipped.  Typing:

`T<` would result in the correct insertion of a matching closing
chevron: `T<|>`, but typing `>` would not result in `T<>|>`, not
skipping the closing chevron even with the `autoskip` action set on the
pair.

This was caused by the regexp in `sp-rust-could-be-parameterized` set as
a `when` filter for the local pair "<" ">".  This filter is called by
`sp--do-action-p` not only when inserting a chevron, but also when
determining if the closing chevron should be autoskipped.  In the
closing case, the regexp does not match, because at the moment this
check is done the code contains one extra chevron: `<T>|>`.

This commits adds a ">" alternative to the regexp so that filter returns
true when `sp--do-action-p` is called for the autoskip action on the
closing chevron.

Closes #642.

### DIFF
--- a/smartparens-rust.el
+++ b/smartparens-rust.el
@@ -62,15 +62,30 @@ If we return nil, ' should be used for character literals."
             (goto-char paren-pos)
             (looking-at "<"))))))
 
-(defun sp-rust-could-be-parameterized (&rest args)
-  "Return t if we could add a <T> in this position.
-If nil, the user is probably using < for something else."
-  (and (apply #'sp-in-code-p args)
-       (looking-back (rx (or letter (seq letter "<") (seq letter "::<"))))))
+(defun sp-rust-filter-angle-brackets (id action context)
+  "Return t if we should allow the ACTION in the current CONTEXT
+  for angle brackets."
+  (cond
+   ;; Disallow in format string after any character, since < is used to specify
+   ;; alignment
+   ((and (eq context 'string)
+         (looking-back (rx (seq alphanumeric "<")))) nil)
+
+   ;; Disallow when inserting in code in two situations: 1) when '<' is used for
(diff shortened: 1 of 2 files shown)


fmdkdd:
@Fuco1 It's fixed for me.  But I think I want to add a test to ensure this stays fixed in the future.

fmdkdd:
I've added a test for this particular case and amended the commit.  Should be good to go.

Fuco1:
Another solution would be check the `action` parameter for `'insert` and only do the check then.  Is the added code useful in all the variants?

In the new system you will have an option to add predicates for each action separately which will really help the reuse... now you need to switch/case/cond each one which is quite ugly.

Anyway I'm happy with any fix :)

fmdkdd:
@Fuco1 Right!  I rewrote the filter to be more explicit.  It still passes all the tests, and I've added a couple to catch this particular situation with autoskipping.

Let me know what you think.

Fuco1:
Great, I have nothing to add :) Thank you.

fmdkdd:
Thank you for your help :)

Issue #649, opened 2016-09-13
backtick and single quote should not be pairs in racket-mode


jgarvin:
Title pretty much says it all, like other lisps they are used for macros and quoting, not for delimiting anything.

xiongtx:
This also occurs with docstrings in Clojure. A backtick is matched with a single quote, which is the right thing to do in Emacs Lisp but not Clojure. Instead, a backtick should be matched with another backtick.

Where is this behavior controlled in source?

Fuco1:
https://github.com/Fuco1/smartparens/blob/master/smartparens-config.el#L66

You would need to write a new rule for clojure in `smartparens-clojure.el`

Fuco1:
@jgarvin looking at `smartparens-racket.el` it seems that backtick is already disabled.  So what should be the correct behaviour?

Fuco1:
Seems to be fixed in `smartparens-racket`.

PR #659, opened 2016-10-14
Match "`" with "`" in Clojure-related modes

xiongtx:
Addresses #649 for Clojure-related modes.

### DIFF
--- a/smartparens-clojure.el
+++ b/smartparens-clojure.el
@@ -45,5 +45,12 @@
 (dolist (mode '(clojure-mode clojurescript-mode clojurec-mode cider-repl-mode))
   (add-to-list 'sp-sexp-prefix `(,mode regexp ,sp-clojure-prefix)))
 
+;; Match "`" with "`" in strings and comments
+(sp-with-modes sp-clojure-modes
+  (sp-local-pair "`" "`"
+                 :when '(sp-in-string-p
+                         sp-in-comment-p)
+                 :unless '(sp-lisp-invalid-hyperlink-p)))
+
 (provide 'smartparens-clojure)
 ;;; smartparens-clojure.el ends here
--- a/smartparens-config.el
+++ b/smartparens-config.el
@@ -65,7 +65,9 @@
 ;; lisp modes too
 (sp-with-modes sp-lisp-modes
   ;; disable ', it's the quote character!
-  (sp-local-pair "'" nil :actions nil)
+  (sp-local-pair "'" nil :actions nil))
+
+(sp-with-modes (cl-set-difference sp-lisp-modes sp-clojure-modes)
   ;; also only use the pseudo-quote inside strings where it serve as
   ;; hyperlink.
   (sp-local-pair "`" "'"
--- a/smartparens-config.el
+++ b/smartparens-config.el
@@ -67,7 +67,7 @@
   ;; disable ', it's the quote character!
   (sp-local-pair "'" nil :actions nil))
 
-(sp-with-modes (cl-set-difference sp-lisp-modes sp-clojure-modes)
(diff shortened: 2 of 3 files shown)


smartparens.el
```diff
@@ -516,7 +516,19 @@ Symbol is defined as a chunk of text recognized by
                            slime-repl-mode
                            stumpwm-mode
                            )
-  "List of Lisp modes."
+  "List of Lisp-related modes."
+  :type '(repeat symbol)
+  :group 'smartparens)
+
+(defcustom sp-clojure-modes '(
```
Fuco1: This should probably be moved into the clojure file.

smartparens-config.el
```diff
@@ -65,7 +65,9 @@
 ;; lisp modes too
 (sp-with-modes sp-lisp-modes
   ;; disable ', it's the quote character!
-  (sp-local-pair "'" nil :actions nil)
+  (sp-local-pair "'" nil :actions nil))
+
+(sp-with-modes (cl-set-difference sp-lisp-modes sp-clojure-modes)
```
Fuco1: Use `-difference` instead of `cl-set-difference`.  It's just a style preference, but we use dash over cl.

smartparens-clojure.el
```diff
@@ -45,5 +45,12 @@
 (dolist (mode '(clojure-mode clojurescript-mode clojurec-mode cider-repl-mode))
   (add-to-list 'sp-sexp-prefix `(,mode regexp ,sp-clojure-prefix)))
 
+;; Match "`" with "`" in strings and comments
+(sp-with-modes sp-clojure-modes
+  (sp-local-pair "`" "`"
+                 :when '(sp-in-string-p
+                         sp-in-comment-p)
+                 :unless '(sp-lisp-invalid-hyperlink-p)))
```
Fuco1: I don't know if in clojure you can use the grave accent for pseudoquoting, but if so you would also need the `:skip-match` predicate.  That will make sure the pair is not parsed in context where it should not be parsed.

smartparens.el
xiongtx: If we move `sp-clojure-modes` into the `smartparens-clojure.el` file, can we still use `sp-clojure-modes` in `smartparens.el`? I don't see a `(require 'smartparens-clojure')` in `smartparens.el`.

smartparens-clojure.el
xiongtx: Clojure does have syntax-quote, but I don't know why we'd need the `:skip-match`:

```emacs-lisp
(lambda (ms mb me)
  (cond
   ((equal ms "'")
    (or (sp-lisp-invalid-hyperlink-p "`" 'navigate '_)
        (not (sp-point-in-string-or-comment))))
   (t (not (sp-point-in-string-or-comment)))))
```

We are already specifying `:when '(sp-in-string-p sp-in-comment-p)`; why do we need `sp-point-in-string-or-comment`?

Fuco1: You're right, this is to handle the straight quote when it is not part of the hyperlink... (for example `it's`) but in clojure this is not a problem.  I got confused :P

xiongtx: :+1:

smartparens.el
xiongtx: Yes, we to exclude Clojure modes from Lisp modes that match backtick with single quote.

In fact, it seems that other modes (e.g. Racket mode from #649) should also be excluded.

Fuco1: You have a point.  We started to lump them all together maybe too hastily.  That variable still controls a lot of behaviour, like the single quotes or slurping (and other match filters).

Still, it doesn't feel right to have this in the main file.  It is the only solution for now though, so let's leave it like it is now.

Fuco1:
Thanks!

PR #670, opened 2016-11-04
Fix whitespace insertion when slurping in LaTeX

PythonNut:
See https://github.com/Fuco1/smartparens/pull/631#issuecomment-258503987


Fuco1:
Thanks!

PR #673, opened 2016-11-08
some nice scala defaults

fommil:
this is a WIP because I'm not sure how to write the tests correctly:

- `execute-kbd-macro` is inserting text explicitly, and not triggering the special rules
- I don't know how to test the region wrapper. Do you have syntax for that or should I use `set-mark`?

### DIFF
--- a/Cask
+++ b/Cask
@@ -14,6 +14,7 @@
   (depends-on "shut-up")
   (depends-on "ruby-mode")
   (depends-on "racket-mode")
+  (depends-on "scala-mode")
   (depends-on "rust-mode")
   (depends-on "auctex")
   (depends-on "clojure-mode")
--- a/smartparens-scala.el
+++ b/smartparens-scala.el
@@ -53,5 +53,27 @@
 (--each '(scala-mode inferior-scala-mode)
   (add-to-list 'sp-sexp-suffix (list it 'regexp "")))
 
+(defun sp-scala-wrap-with-indented-newlines (_ action _)
+  "Post handler for the wrap ACTION, putting the region on indented newlines."
+  (when (eq action 'wrap)
+    (sp-get sp-last-wrapped-region
+      (let ((beg :beg-in)
+            (end :end-in))
+        (save-excursion
+          (goto-char end)
+          (newline-and-indent))
+        (save-excursion
+          (goto-char beg)
+          (newline))
+        (indent-region beg end)))))
+
+(sp-local-pair 'scala-mode "(" nil
+               :post-handlers '(("||\n[i]" "RET")))
+
+(sp-local-pair 'scala-mode "{" nil
+               :post-handlers '(("||\n[i]" "RET")
+                                ("| " "SPC")
+                                sp-scala-wrap-with-indented-newlines))
+
 (provide 'smartparens-scala)
 ;;; smartparens-scala.el ends here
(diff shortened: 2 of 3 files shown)


Fuco1:
`execute-kbd-macro` should trigger everything the same way as if you were typing it, we use it all over in the tests.  If it doesn't work there might be some problem with the setup.

As for tests, 99% can be covered with `sp-test-with-temp-buffer`, check the docstring (it's in `test/test-helper.el` and just grep the test folder for usage.  Tldr `|` in the input represents point and `M` (capital m) represents mark...

Fuco1:
With `execute-kbd-macro` don't forget to wrap with `kbd` if you use special syntax, for example

    (execute-kbd-macro (kbd "DEL"))

Would excute "delete" key but

    (execute-kbd-macro "DEL")

would just write the letters `DEL` literally.

fommil:
of course :smile: I'll update later tonight when I get home from work. Helps to keep my mind of the world that is burning around us.

fommil:
I updated the PR but I don't think that changes the way the `execute-kbd-macro` works. e.g. `(kbd "SPC")` just returns `" "`.

Fuco1:
The problem might be with the conditional handlers... I'm not sure we have tests for that yet.  I'll try to look at that as well.

fommil:
I also completely forgot about this.

fommil:
hmm, I never worked out what was going on here. AFAIK it should be passing ... did you ever find the time to look into what the problem could be?

Fuco1:
I'm going to right after your other PR.  I feel terrible now having it lying here for so long.  I'm on vacation now which is the best time for hacking! Thanks for your patience.

fommil:
it's ok I think I have these in my personal config so I haven't been missing out :smile: the ability to hack my own emacs sometimes makes me lazy at implementing features properly in my own projects and then I talk to people who ask "can you make it do X?" and I'm like "it already does X, what do you mean? oooooh... that one is in my `.emacs.d/lisp` folder"

Fuco1:
I feel the same way about my flycheck hacks :/ Still, I've read a nice blog about [optimistic merging](http://hintjens.com/blog:106) recently and I would like to start following that.  It engages people a lot better.

Fuco1:
Turns out `kbd` and `execute-kbd-macro` are quite picky.

You need to execute the entire thing in one call to `execute-kbd-macro` otherwise the events won't come to the handler as during the real execution (probably some edebug or other command gets executed inbetween breaking the simulation).  Also, literal space characters are *removed* from `kbd` spec and all the special chars delimited by spaces are translated.  So you end up with something like

```elisp
(execute-kbd-macro (kbd "{ RET f SPC => SPC f.toString"))
```

which "types" `{\n  f => f.toString`

I had to dig a bit into Emacs to get to the bottom of this.  Reading code always works :D

fommil:
Wow, that's really bizarre. We should add this to a comment to save some poor soul from suffering in the future.

Fuco1:
Closed in 10657cc.

Thanks for the contribution!

Fuco1:
> We should add this to a comment to save some poor soul from suffering in the future.

Good point, I will add it to the testing page on the wiki as well.

Fuco1:
https://github.com/Fuco1/smartparens/wiki/Testing#simulating-user-typing

fommil:
Thank you again! Need to find myself some time to hack on some Emacs lisp soon. Been missing it.

PR #677, opened 2016-11-20
Correct use of eval-after-load

Silex:
With a symbol, only the file providing the feature triggers the
eval-after-load. With a string, any file named the same way triggers
it. This is problematic for custom configurations named after
packages.


Fuco1:
Did you check all the packages and the features they provide? I'm not sure if it is named the same in each case.

Silex:
Hum, I didn't think of that possibility! I'll try to check

Silex:
Ok I verified, and I also corrected things I missed in the other forms :-)

Fuco1:
There is `(provide 'tex-mode)` which defines `latex-mode` but there is also `auctex` which has a file `latex.el` which provides `latex` feature (it also provides `tex` and `tex-site` and gods know what else... nobody understands how that loads :D

Hooking on `tex-mode` and `latex` should be enough I think.

Silex:
A'ight

PR #689, opened 2016-12-30
Test against 25.1 and trunk too

Wilfred:


Fuco1:
I have this prepared on a branch but had constant problems with evm/cask not being ready for 25.1. Seems it's still not ready as your build fails on some cask error ("cannot be bootstrap").

Wilfred:
OK, second attempt. I wanted to create a PR to make sure I didn't break master. I use Cask and EVM on a number of other elisp projects, so this should be doable.

Fuco1:
Check my branch too then, I have some simplifications there, maybe you can get it to work together.

Wilfred:
OK, I think this is now building correctly. I had the wrong Emacs name `EVM_EMACS=emacs-24.4-bin` should be `EVM_EMACS=emacs-24.4-travis` with the latest EVM and container-based Travis.

The build is failing on 25.1, but that looks legitimate.

Wilfred:
OK, merged. It would be nice to fix the tests on 25.1, but this gets them running.

Fuco1:
All right, I'll go over the tests and poke around a bit. Thanks for the setup.

PR #691, opened 2016-12-31
Support < and > in Rust with strict mode.

Wilfred:
Ensure that we can insert > in function signatures, or delete it in
comparisons and left shifts.

@Fuco1 I'm gradually learning the smartparens API, but handling all the different actions is tricky. I think this is correct, but I'd appreciate your review.

If you're too busy to review, I'll merge in a few days anyway, no stress.

### DIFF
--- a/smartparens-rust.el
+++ b/smartparens-rust.el
@@ -63,29 +63,51 @@ If we return nil, ' should be used for character literals."
             (looking-at "<"))))))
 
 (defun sp-rust-filter-angle-brackets (id action context)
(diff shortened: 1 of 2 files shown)


Fuco1:
This looks good.  I suppose it doesn't break the previous setup (was it covered by tests?).  If it wasn't, maybe add some tests to cover the old behaviour as well so we don't regress (optional).

Wilfred:
Thanks. 

We have reasonable test coverage, and I've added tests for the cases that this PR fixes.

Issue #690, opened 2016-12-30
sp-backward-delete-char mistakes end of line in comint buffers as end of buffer


dwcoates:
sp-backward-delete-char seems to not work at end of line in comint buffers. It thinks it's at the end of the buffer when at the end of the line. 

`C-e sp-backward-delete-char` yields `End of buffer`

Occasionally it also hangs. Some `C-g` spam successfully quits

Fuco1:
Thanks for the report.  Comint is our eternal enemy, it does *weird* shit :P

dwcoates:
I still love it, though <3

PR #694, opened 2017-01-15
Fix backward-delete hang in comint when at end of last prompt

dwcoates:
This should be a fix for #690. It at least stops the "end of buffer" problem... I'll have to keep testing to see if this has anything to do with the hanging that `C-g` was killing (that might be a jupyter ipython problem, because I haven't gotten it since switching to normal python repl).

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -8228,7 +8228,9 @@ The return value is actually cons pair of opening and closing
 string delimiter enclosing this string."
   (setq pos (or pos (point)))
   (when (and (sp-point-in-string)
-             (save-excursion (forward-char) (not (sp-point-in-string)))
+             (save-excursion (if (eq (point-max) (point))
+                                 t
+                               (forward-char) (not (sp-point-in-string))))
              (save-excursion (backward-char) (not (sp-point-in-string))))
     (save-excursion
       (let* ((syntax (nth 3 (syntax-ppss pos)))
--- a/smartparens.el
+++ b/smartparens.el
@@ -8228,7 +8228,7 @@ The return value is actually cons pair of opening and closing
 string delimiter enclosing this string."
   (setq pos (or pos (point)))
   (when (and (sp-point-in-string)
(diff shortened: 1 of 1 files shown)


Fuco1:
First of all, thanks for looking into this!  Now let's analyze a bit.

The situation you are fixing there should be this: I am inside an unclosed empty string with the point at the end of the buffer.  Which means the prompt looks something like

    > foo "|

If it isn't then the code probably need to change elsewhere and this is just some side-effect.

The problem with comints is that they often produce invalid output, i.e. in the output of the command there might be quotes which taken together with the "repl" code will create a messed up parse.

The general solution I have in mind is to always narrow the buffer to just the prompt or just the area between prompts and ignore the rest for the purpose of balance checks.  The problem is that narrowing doesn't flip the string states and so you would need to copy to a new buffer and there things get rather messy.

dwcoates:
Hmm. I figured the problem was sp-backward-delete-char attempting to forward-char at eof, and because forward-char raises an error at eof, it was uncaught and simply failing the command (resulting in the "End of buffer" in message prompt). Now I remember, though! Despite actually referencing the issue #, I've conflated two problems (should have made two different issues for them, admittedly): 1) sp-backward-delete-char simply failing with error message when at end of last comint buffer, which I believe this solves (but I'll have to test a bit more). and 2) sp-backward-delete-char hanging indefinitely when confronted with an (unmatched?) closed paren at end of coming buffer, which I suspect this does not solve. I can't replicate the latter, but I'll keep trying.

dwcoates:
I have yet to have any hanging at eof in iPython, iESS, or Python repl buffers. Also, no end-of-buffer errors. My iPython buffer has native completions off, so it is not using the ipython interpreter instance for completions or anything. Probably irrelevant, but I have no idea how comint works...

dwcoates:
Btw, this is a fix for both problems **(1)** and **(2)** described above.

Fuco1:
All right, if it works it works :D Let's merge.

PR #700, opened 2017-01-29
Added some missing special forms for the Elixir mode

gausby:
The code that check for code blocks missed some special forms. I think I got them all.


Fuco1:
Amazing, thanks!

PR #704, opened 2017-02-02
Add GitHub issues template

xiongtx:
An issue template should cut down on the number of incomplete bug reports.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,22 @@
+## Expected behavior
+
+
+
+## Actual behavior
+
+
+
+## Steps to reproduce the problem
+
+
+
+## Backtraces if necessary (`M-x toggle-debug-on-error`)
+
+
+
+## Environment & version information
+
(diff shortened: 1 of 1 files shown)


xiongtx:
Ping @Fuco1 🙂

Fuco1:
@xiongtx oh dear, I'm terribly sorry! I started the review with the comment above but never "submitted" it :D Apparently github requires you to press the "send" button twice.  That was 18 days ago so I assumed this one was on you :D

Please give your opinion on the review and then we can move forward!

xiongtx:
I added "Active major mode" to the template.

I'd disagree that this template is useful only for bug reports. The "expected behavior--actual behavior" format is equally useful for feature requests. 

You wouldn't submit a stacktrace for a feature request, of course, but things like `smartparens` version remain important. It is not uncommon for users to request a feature that already exists in the latest build, and for the confusion to clear up only after revealing that they're using an older version of the package.

Fuco1:
That's true, thanks for that point.  Good to merge on my side :)

PR #712, opened 2017-02-25
Allow inserting closing delimiters for non-autoskip pairs

johnmastro:
Don't inhibit inserting closing delimiters for pairs that don't use the `autoskip` action.

Commit a677859 added code to prevent closing delimiters from being inserted in strict mode. Typing a closing delimiter will either skip over the delimiter that was automatically inserted, or emit a message that an unbalanced delimiter was blocked from insertion. I think it makes sense to exempt from this pairs that don't use the `autoskip` action (and/or the `insert` action? more on that below), since in that case the only reason to type the closing delimiter is if you want to insert it.

I ran into this because, in C-like modes, I set up `{}` for the `navigate` and `wrap` actions, but not `insert` or `autoskip`.

It may also make sense to exempt pairs without the `insert` action, for basically the same reason - if you've configured Smartparens not to insert the closing delimiter for you, you probably intend to type it yourself. However, I never use `insert` without `autoskip` or _vice versa_, so I'm not sure. Of the two, I went with `autoskip` because it is more directly connected to typing a closing delimiter.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -3694,11 +3694,13 @@ active region."
     (--any? (string-match-p (regexp-quote char) it))))
 
 (defun sp--char-is-part-of-closing (char &optional pair-list)
-  "Return non-nil if CHAR is part of a pair delimiter of length 1."
-  (->> (or pair-list (sp--get-pair-list))
-    (--filter (= 1 (length (cdr it))))
-    (-map 'cdr)
-    (--any? (string-match-p (regexp-quote char) it))))
+  "Return non-nil if CHAR is part of a pair delimiter of length 1.
+Specifically, return the pair for which CHAR is the closing
+delimiter."
+  (let ((regexp (regexp-quote char)))
+    (->> (or pair-list (sp--get-pair-list))
+         (--filter (= 1 (length (cdr it))))
+         (--find (string-match-p regexp (cdr it))))))
 
 ;; TODO: this only supports single-char delimiters.  Maybe it should
 ;; that that way.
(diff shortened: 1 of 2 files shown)


johnmastro:
For a test, would something like this be appropriate?

```emacs-lisp
(ert-deftest sp-test-sp-dont-inhibit-closing-pair-unless-autoskip ()
  (sp-test-with-temp-buffer "" nil
    (let ((sp-pairs sp-pairs))
      (smartparens-strict-mode)
      (sp-local-pair major-mode "{" "}" :actions '(insert autoskip navigate wrap))
      (execute-kbd-macro "}")
      (should (eq (char-before) nil))
      (sp-local-pair major-mode "{" "}" :actions '(:rem autoskip))
      (execute-kbd-macro "}")
      (should (eq (char-before) ?\})))))
```

smartparens.el
```diff
@@ -3833,14 +3835,16 @@ achieve this by using `sp-pair' or `sp-local-pair' with
...
+autoskip, and we can not jump out of its enclosing sexp (i.e. it
+does not match), we are not allowed to insert it literally
+because it would break the balance; so we delete the
+just-inserted character."
   (when (and smartparens-strict-mode
-             (sp--char-is-part-of-closing
-              last (sp--get-allowed-pair-list))
+             (-when-let (pair (sp--char-is-part-of-closing
+                               last (sp--get-allowed-pair-list)))
+               (memq 'autoskip (plist-get (sp-get-pair (car pair)) :actions)))
```
Fuco1: We can pass the keyword as an argument to `sp-get-pair` which I think is slightly better if we later decide to change the internal representation.

Fuco1:
The test is fine if it works :)

johnmastro:
Thanks for taking a look and for the feedback. I force-pushed an updated commit using the optional argument for `sp-get-pair` I had missed plus adding the test.

Fuco1:
Great! Sorry it took me so long.

In the future don't hesitate to @ ping me if I take too long.  My github workflow suck and I get tasks lost so often :/

johnmastro:
No need to apologize but I will ping in the future - didn't want to be a bother if you just hadn't had a chance to get to it :) Thanks for reviewing/merging!

PR #714, opened 2017-02-26
Support fsharp-mode

juergenhoetzel:
Rename smartparens-ocaml to smartparens-ml: The OCaml local pair
definitions apply for all ML based languages (like F#).


juergenhoetzel:
F# is basically an OCaml dialect on the CLR. The same problems discussed in 
[forward/backward in OCaml](https://github.com/Fuco1/smartparens/pull/631) also apply in `fsharp-mode`.
It would be nice to also have [F#/OCaml mode pairs](https://github.com/fsharp/emacs-fsharp-mode/issues/100) in `fsharp-mode`

Fuco1:
One thing that slightly bothers me is that if someone required the file directly their config will break now.  But I think that's a minimal number of people and if we put a note large enough in the release notes it should be fine.

Thanks for the contribution!

PR #715, opened 2017-03-01
Update smartparens-rust.el

ljos:


Fuco1:
Oh the perils of copy paste :D Thanks for catching that :)

PR #716, opened 2017-03-03
JavaScript support

marinintim:
Hello! I want to support JavaScript in smartparens.

This pull request adds smartparens-javascript, which I plan to expand later. Right now, it fixes issue with creating spaces around dots.

### DIFF
--- a/smartparens-config.el
+++ b/smartparens-config.el
@@ -117,6 +117,8 @@
 (eval-after-load 'tex-mode                 '(require 'smartparens-latex))
 (eval-after-load 'tuareg                   '(require 'smartparens-ml))
 (eval-after-load 'fsharp-mode              '(require 'smartparens-ml))
+(--each '(javascript-mode js2-mode)
+  (eval-after-load it                      '(require 'smartparens-javascript)))
 (provide 'smartparens-config)
 
 ;;; smartparens-config.el ends here
--- a/smartparens-javascript.el
+++ b/smartparens-javascript.el
@@ -1,0 +1,55 @@
+;;; smartparens-javascript.el --- Additional configuration for JavaScript based modes.
+
+;; Copyright (c) 2017 Marinin Tim
+;; Author: Tim Marinin <mt@marinin.xyz>
+;; Maintainer: Tim Marinin <mt@marinin.xyz>
+;; Created: 2017-03-03
+;; Keywords: abbrev convenience editing javascript
+;; URL: https://github.com/Fuco1/smartparens
+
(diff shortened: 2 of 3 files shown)


Fuco1:
Hi, thanks for taking the time to do this!

Could I bother you to add a test case which "confirms" the change?  So that we are sure not to break this later on.  You can look in `test/smartparens-<LANGUAGE>-test.el` files to get some inspiration.  The python and rust ones are pretty clean.

marinintim:
@Fuco1: hi, I've added a test file, do I need to add it somewhere else?

Fuco1:
You placed it correctly, you can see it being executed on Travis.  But the code doesn't seem to work for some reason :O (the test fails correctly because the result is incorrect).

Is javascript-mode built-in? If not, you would need to add it to `Cask` as well to get it installed during the test setup.

marinintim:
Fuco1: could you check smartparens-javascript.el? It looks like my code is not applied for this temporary buffer in tests. If I execute code manually from smartparens-javascript.el, I see it working in *scratch* buffer with javascript-mode. Also, javascript-mode is a built-in mode.

I'm new to Elisp, so I could easily mess something up.

Fuco1:
It seems `javascript-mode` is aliased to `js-mode`... in some versions of Emacs, and so when I compare the mode symbol returned by `major-mode` they don't match your list.  Try adding `js-mode` to the list of mode where we apply this change.  That should work.

marinintim:
@Fuco1: could you take a look at test results? It seems that at some point it stalled in unrelated place, but I could miss something. I've added `js-mode` to the list of modes.

Fuco1:
I have just experienced the same issue working on a different branch.  I've reported it back to undercover: https://github.com/sviridov/undercover.el/issues/25

I guess we'll have to wait now :/

marinintim:
@Fuco1: Hi, is there any updates on undercover issue?

Fuco1:
No, and I'm starting to have problems in other projects as well.  I think I will disable it for now, this PR is good to go.

marinintim:
@Fuco1: so you would merge it now?

Fuco1:
@marinintim I've rebased and cleaned up the history, it is now on master.

Thanks for the contribution!

yqrashawn:
Hi @marinintim, what's your plan for js with sp. I think there need to be a `javascript-sexp` distinguish by `{...}`, `[ ( or , ]{...}[ ) or , ]` for fast navigation, and normal sexp for string manipulation.

Issue #724, opened 2017-03-30
Smartparens and Evil Registers


jojojames:
## Expected behavior

When using Smartparens, yank/kill actions puts the text into the right register.

## Actual behavior

Smartparens doesn't try to do anything with evil registers.

## Steps to reproduce the problem

sp-kill-sexp's text should have something in register 1 on evil.

sp-copy-sexp's text should have something in register 0 on evil.

I have some changes to make Smartparens work better with Evil.

They do two things.
1. Anything that 'copies' gets put into the 0th register to match vim's convention.
2. Evil/Vim, you can prefix a register before a command to have that command's kill/yank/paste(whatever) go into that register after that command.

This should let you do something like bind sp-copy-sexp to M-y and then do M-y on a sexp.

Result:

For M-y, 0 register should be populated.
And for "a M-y, the 0 and a register should be populated.
```
(eval-after-load 'evil
    (lambda ()
      (defun jojo/sp-copy-sexp-to-0-register (&rest _args)
        "After `sp-copy-sexp', put copied text into 0 register like vim."
        (interactive)
        (evil-set-register ?0 (evil-get-register ?1))
        (when evil-this-register
          (evil-set-register evil-this-register (evil-get-register ?0))
          (setq evil-this-register nil)))
      (advice-add 'sp-copy-sexp :after #'jojo/sp-copy-sexp-to-0-register)
      (advice-add 'sp-backward-copy-sexp :after #'jojo/sp-copy-sexp-to-0-register)

      (defun jojo/sp-delete-use-register (&rest _args)
        "After some `smartparens' kill commands, add the register to evil."
        (interactive)
        (when evil-this-register
          (evil-set-register evil-this-register (evil-get-register ?1))
          (setq evil-this-register nil)))

      (advice-add 'sp-kill-sexp :after #'jojo/sp-delete-use-register)
      (advice-add 'sp-backward-kill-sexp :after #'jojo/sp-delete-use-register)))
```

It would be great if this functionality can go into smartparens proper instead of using advices.

Fuco1:
I bind `(kill-fn (if dont-kill 'copy-region-as-kill 'kill-region))` in `sp-kill-sexp`, this function is responsible for all the later killing/copying.

[...]

PR #725, opened 2017-04-01
Replace kill-fn with wrapper that accounts for Evil Yank/Registers

jojojames:
Evil/Vim uses the 0 Register as the Yank Register. When yanking a
sexp, add the yanked text to the 0 Register too.

Added logic to add copied region to 0 Register like in Vim/Evil.

Added Register Prefix handling so that the yanked text goes into the
right place.

For example, this should be possible now.

"x (sp-copy-sexp)

:reg

0 Register and x register should contain the text that was copied.

https://github.com/Fuco1/smartparens/issues/724

Open to ideas on the Doc comment for the new function.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -5978,6 +5978,24 @@ Examples:
 If the next command is `sp-kill-sexp', append the whitespace
 between the successive kills.")
 
+(defun sp--kill-or-copy-region (beg end &optional dont-kill)
+  "Kill or copy region between BEG and END according to DONT-KILL.
+If `evil-mode' is active, copying a region will also add it to the 0 register.
+Additionally, if command was prefixed with a register, copy the region
+to that register."
+  (interactive)
+  (let ((result
+         (if dont-kill
+             (copy-region-as-kill beg end)
+           (kill-region beg end))))
+    (when (bound-and-true-p evil-mode)
+      (when dont-kill
+        (evil-set-register ?0 (evil-get-register ?1)))
+      (when evil-this-register
+        (evil-set-register evil-this-register (evil-get-register ?1))
+        (setq evil-this-register nil)))
+    result))
+
 (defun sp-kill-sexp (&optional arg dont-kill)
   "Kill the balanced expression following point.
 
@@ -6040,8 +6058,7 @@ Note: prefix argument is shown after the example in
          (n (abs arg))
          (ok t)
          (b (point-max))
-         (e (point))
-         (kill-fn (if dont-kill 'copy-region-as-kill 'kill-region)))
+         (e (point)))
     (cond
      ;; kill to the end or beginning of list
      ((and raw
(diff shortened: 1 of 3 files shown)


Fuco1:
Continuation of #724.

smartparens.el
```diff
@@ -6047,21 +6061,25 @@ Note: prefix argument is shown after the example in
               (let ((del (sp-get-whitespace)))
                 (sp-get del (delete-region :beg :end))))
           (if (> arg 0)
-              (funcall kill-fn (sp-get next :beg-prf) (sp-get enc :end-in))
-            (funcall kill-fn (sp-get next :end) (sp-get enc :beg-in)))
+              (sp-copy-or-kill-region
+               dont-kill (sp-get next :beg-prf) (sp-get enc :end-in))
```
Fuco1: I think that the `dont-kill` could come as 3rd optional argument.  If we later decide to use this function for other killing functions in SP, they all *do* kill so we would have to pass a "useless" nil always as first argument.

Maybe we could name it `kill-or-copy` instead so the first "default" verb comes first.

Fuco1:
Could you also add some simple tests to see if the registers get filled?  If you don't know where to start, there's an article on the wiki about testing SP.

smartparens.el
```diff
@@ -5972,6 +5972,21 @@ Examples:
 If the next command is `sp-kill-sexp', append the whitespace
 between the successive kills.")
 
+(defun sp-copy-or-kill-region (dont-kill beg end &optional region)
```
Fuco1: Oh, I see you mentioned the docstring in your post, sorry for missing that :)

Just say something like "KIll or copy region between BEG and END according to DONT-KILL" and then explain the extra `evil` behaviour, "If evil-mode is active... it does this and that".

jojojames: Will do on the docstring and --.

kill-region and copy-region-as-kill both take an optional region parameter. I included that to match the internal calls in case we do end up using the region.

smartparens.el
jojojames: Good idea. Will do.

jojojames:
^ Sure thing.

smartparens.el
Fuco1: Hmm, reading the docstring on `kill-region`

    The optional argument REGION if non-nil, indicates that we're not just killing
    some text between BEG and END, but we're killing the region.

I have no idea what that means :D AFAIK the text between BEG and END is the region?

Looking at the code, it seems to be only used as some sort of a flag.  I don't know, I would remove it and then add it later if we need it.

jojojames:
Hmnn, trying to run the tests get me this.

```
cask exec ert-runner
Wrong number of arguments: (3 . 3), 2
```

I'll have to see if I can figure that out.

smartparens.el
```diff
@@ -5972,6 +5972,24 @@ Examples:
 If the next command is `sp-kill-sexp', append the whitespace
 between the successive kills.")
 
+(defun sp--kill-or-copy-region (beg end dont-kill)
```
Fuco1: I would make `dont-kill` optional here, as discussed before.

Fuco1:
About the tests, I figured it might be that you don't pass the argument somewhere, or that there is a "shifted" paren (so that only two args instead of three are getting in), but I haven't found any such instance.

Is there any backtrace?

smartparens.el
jojojames: Whoops, missed that. Will do.

jojojames:
I run the tests from the terminal.

```
james:smartparens/ (master) $ pwd
/Users/james/.emacs.d/fork/smartparens
james:smartparens/ (master) $ cask exec ert-runner
Wrong number of arguments: (3 . 3), 2
james:smartparens/ (master) $ 

james:smartparens/ (master) $ cask --version
0.8.1
james:smartparens/ (master) $ which cask
/usr/local/bin/cask
james:smartparens/ (master) $ 
```

jojojames:
Looks like the ert runner doesn't work for me on OSX.

I wrote and ran those tests on Linux with no hitch.

Fuco1:
@jojojames Hey, I think I've fixed the builds on master.  If you could squash & rebase on top, we let the CI run and then I think this is good to go!

jojojames:
Done.

Fuco1:
Awesome, merged!

PR #727, opened 2017-04-03
add `defp` to elixir config

boxxxie:


Fuco1:
Thanks, amazing!  Sorry for the delay, I somehow lost track of this.

If you contribute in the future don't be afraid to bug me! :)

boxxxie:
There is a problem with kill line when using single line do syntax. Got a clue for me on how to tackle it? I still don't understand the elixir mode code very well

Fuco1:
The skip function in smartparens doesn't recognize the single line syntax, there is some issue opened for that already.  The function in question is `sp-elixir-def-p`.  If the `do` is part of the definition it is not parsed as a pair, instead the "opening" keyword is considered the opening delimiter.  The check doesn't work with the single line syntax.

PR #740, opened 2017-04-29
update local pairs for tuareg-mode and reason-mode

### LINKED ISSUE #730: Delay in opening a comment by '(* *)' in big OCaml file when smartparens is enabled.
I just want to follow up with this issue is that starting a comment in a big size OCaml file can be slow in the default setting, i.e., the pair (* *) isn't set as a local pair in smartparens.

## Expected behavior

- In OCaml, a comment is enclosed by a pair of `(*` and `*)`.

- When smartparens is enable, opening a comment in big OCaml file should work smoothly.

## Actual behavior

- When smartparens is enable, If I type `(*` to open a comment in a big OCaml file (> 2000 lines of code), the response from Emacs become very slow.

- Enable smartparens
- Open a big OCaml file, for example: https://github.com/ocaml/merlin/blob/master/src/utils/menhirLib.ml
- Go to a random line, i.e line 100, type `(` first, then smartparens automatically makes it become `()`.
- Now continue to type `*` in the middle of () to obtain (*|), where | is the cursor position,
then Emacs becomes very slow. 
- I guess it tries to re-parse the file and is problematic somewhere.

- `smartparens` version: 2017-401.315, as shown after being installed in elpa
- Active major-mode: tuareg
- Emacs version (`M-x emacs-version`): 25.1
- OS: Linux Mint 18.1

tddsg:
Hi,

I updated local pairs for 2 modes: tuareg-mode and reason-mode in the `smartparens-ml` package. This update will fix the issue described in https://github.com/Fuco1/smartparens/issues/730 since the pair `(* *)` will be automatically inserted when typing `(*`.

Could you double check and advise on the changes?

(Note that I change the description since I was using different github account to create the first version of `smartparens-ocaml`).

Thanks.

Note: this fixes #730 when merged.

### DIFF
--- a/smartparens-ml.el
+++ b/smartparens-ml.el
@@ -1,12 +1,9 @@
 ;;; smartparens-ml.el --- Additional configuration for ML languages
 
-;; Copyright (C) 2016 Ta Quang Trung
-
-;; Author: Ta Quang Trung <taquangtrungvn@gmail.com>
-;; Maintainer: Ta Quang Trung <taquangtrung@gmail.com>
+;; Author: tddsg
 ;; Created: 14 July 2016
-;; Keywords: abbrev convenience editing
-;; URL: https://github.com/taquangtrung/smartparens
+;; Updated: 29 April 2017
+;; Keywords: smartparens, ML, ocaml, reason
 
 ;; This file is not part of GNU Emacs.
 
--- a/smartparens-ml.el
+++ b/smartparens-ml.el
@@ -47,5 +47,9 @@
 
 (require 'smartparens)
 
+;;; Local pairs for ML-family languages
+(sp-with-modes '(tuareg-mode) (sp-local-pair "(*" "*)" ))
+(sp-with-modes '(reason-mode) (sp-local-pair "/*" "*/" ))
+
 (provide 'smartparens-ml)
 ;;; smartparens-ml.el ends here


smartparens-ml.el
```diff
@@ -1,12 +1,9 @@
...
 
-;; Copyright (C) 2016 Ta Quang Trung
-
-;; Author: Ta Quang Trung <taquangtrungvn@gmail.com>
-;; Maintainer: Ta Quang Trung <taquangtrung@gmail.com>
+;; Author: tddsg
 ;; Created: 14 July 2016
-;; Keywords: abbrev convenience editing
-;; URL: https://github.com/taquangtrung/smartparens
+;; Updated: 29 April 2017
```
Fuco1: This is not necessary I think, since we have thehistory in git.

smartparens-ml.el
```diff
@@ -1,12 +1,9 @@
 ;;; smartparens-ml.el --- Additional configuration for ML languages
 
-;; Copyright (C) 2016 Ta Quang Trung
-
-;; Author: Ta Quang Trung <taquangtrungvn@gmail.com>
-;; Maintainer: Ta Quang Trung <taquangtrung@gmail.com>
+;; Author: tddsg
```
Fuco1: If there are reasons you don't want your name there that's fine, but the name will still be recorded in the repository history and we can't change that, so this change has relativelly little impact.

Whatever way you decide, the copyright notice should stay (taht is the text Copyright plus the year(s) when changes happened)

Fuco1:
Note: this fixes #730 when merged.

smartparens-ml.el
tddsg: Hi, I changed the name and removed the copyright because looking back, what I added into the package was minor, just 2 lines of code :D Please feel free to edit both the code and the description when you merge these changes into the main repository.

smartparens-ml.el
tddsg: Yes, I agree, please help to remove when you merge it.

PR #746, opened 2017-05-09
Fix: sp-kill-hybrid-sexp breaks C-0 C-k

nochiel:
## Expected behavior

[2.1.2 Killing Lines (Emacs Manual)](http://www.gnu.org/software/emacs/manual/html_node/emacs/Killing-by-Lines.html#Killing-by-Lines) :
>  When `C-k` is given a positive argument N, it kills N lines and the newlines that follow them (text on the current line before point is not killed).  With a negative argument `−N`, it kills N lines preceding the current line, together with the text on the current line before point. *`C-k` with an argument of zero kills the text before point on the current line.

`(kill-line &optional ARG)`:
> Kill the rest of the current line; if no nonblanks there, kill thru newline. With prefix argument ARG, kill that many lines from point. Negative arguments kill lines backward. With zero argument, kills the text before point on the current line.

`(sp-kill-hybrid-sexp ARG)` :
> [...] With ARG numeric prefix 0 (zero) just call `kill-line'. [...]

## Actual behavior
With `C-k` bound to `sp-kill-hybrid-sexp`, `C-0 C-k` kills text after point on current line.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -6346,7 +6346,7 @@ Examples:
                         (current-column)))
          (orig-column (current-column)))
     (cond
-     ((= arg 0) (kill-line))
+     ((= arg 0) (kill-line 0))
      ((and raw (= arg 16))
       (let ((hl (sp-get-hybrid-sexp)))
         (sp-get hl (kill-region :beg-prf :end-suf))))


Fuco1:
Hm, well this is a feature.  In smartparens the `0` argument means "fallback".  It is quite unfortunate that `kill-line` uses it for another purpose, I wasn't aware of that.  But we can't go along with your proposed solution because that would break the convention.

In either case it should probably not just call `kill-line 0` because that might destroy the balance, it would need something like `C-- C-u sp-kill-sexp`

From `sp-kill-sexp` doc:

> With raw prefix C-- C-u, kill all the expressions from beginning of current list up until point.

Fuco1:
Maybe we could add what you requested on `C--` instead? (`negative-argument`).

nochiel:
> In either case it should probably not just call kill-line 0 because that might destroy the balance, it would need something like C-- C-u sp-kill-sexp

Ah. My understanding was that `(sp-kill-hybrid-sexp 0)` destroys the balance anyway. At least, whenever I use it, the balance is destroyed.

But yes `C--` is okay.

Fuco1:
@nochiel that is correct, but that is under the assumption of "fallback" to the regular `kill-line`.  If we changed it to behave like `kill-line` I would assume it would be by default "balance preserving" operation.

Well, the numeric argument right now doesn't behave like `kill-line` either, I'm not sure if there is some logical way to add that behaviour.  Because if we decide to do that later having `C--` (which is really `-1`) to do something else would again be confusing.

Ught, this is unfortunate :D

nochiel:
>  Because if we decide to do that later having C-- (which is really -1) to do something else would again be confusing.

Indeed... a conundrum. And you're right: the sp bindings behaviour should be simple and predictable, which they are now. I'll think some more about it. But I can see that my initial proposal wouldn't work within the current model.

Fuco1:
Thanks for the input! If you figure something out let us know!

PR #748, opened 2017-05-24
Add typescript-mode to smartparens-javascript.el

arichiardi:
The fix you suggested worked so I am PR-ing it :smile: 

Thanks a lot!

I will probably do the same for `reason-mode`, which I am maintaining, at some point.

### DIFF
--- a/smartparens-javascript.el
+++ b/smartparens-javascript.el
@@ -48,7 +48,7 @@
 
 ;; (|sys).path.append---the dot should not travel with the closing
 ;; paren
-(--each '(js-mode javascript-mode js2-mode)
+(--each '(js-mode javascript-mode js2-mode typescript-mode)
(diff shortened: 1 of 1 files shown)


Fuco1:
Great, thanks for confirming and adding this!

If you maintain a mode and would like to support further smartparens configuration I'm pretty open with granting collaborator membership so people can manage "their own stuff" here (I usually prefer it to providing config in the modes themselves but it's up to the maintainers).  If you will be interested just ping me and we'll make it happen.

arichiardi:
@Fuco1 yeah I haven't decided about embedding or not yet. It is a good option for me to hijack either `smartparens-ml.el` or `smartparens-javascript.el` for now. I don't see the configuration too far from Javascript/Rust for instance, even if Reason is a dialect of OCaml. In any case, thanks for solving the problem!

Issue #751, opened 2017-06-02
Add sp-backward-delete-symbol


amosbird:
This would be useful to bypass kill ring.

Fuco1:
Do you need this for interactive use or from program?  From elisp you can temporarily let-bind `kill-ring` to preserve its value.

amosbird:
hmm, I'm not aware of that. Could you show me an example?

Fuco1:
This is the consequence of how dynamic binding works in elisp.

```elisp
(let ((kill-ring kill-ring))
  (do whatever))

;; after the `let' form `kill-ring' reverts to its old value
```

amosbird:
oh, that won't work as expected IMO since my system clipboard will also store the kill ring content when `kill-word` happens.

Fuco1:
If it happens through Emacs during the let binding it will revert.

amosbird:
well, I tested this. It bypassed emacs kill ring but ends up in my system's clipboard.
```elisp
  (defun test-kill ()
    (interactive)
    (let* ((kill-ring kill-ring))
      (sp-backward-kill-symbol 1)))
```

Fuco1:
Can you try it with also locally `let`-binding `select-enable-clipboard` to `nil`?

If it will work that way we can make a patch to SP to add the delete variants.  I don't mind much having that included.  Do you want to create a pull request then?

amosbird:
ah, this works. pr #759

PR #759, opened 2017-06-19
Add backword delete word and symbol variant

amosbird:
targeting #751

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -8626,6 +8626,31 @@ backward direction."
   (interactive "p")
   (sp-backward-kill-symbol arg t))
 
+(defun sp-backward-delete-symbol (&optional arg word)
+  "Delete a symbol backward, skipping over any intervening delimiters.
+
+With ARG being positive number N, repeat that many times.
+
+With ARG being Negative number -N, repeat that many times in
+forward direction.
+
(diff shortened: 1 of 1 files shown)


smartparens.el
```diff
@@ -8626,6 +8626,31 @@ backward direction."
...
+  "Delete a symbol backward, skipping over any intervening delimiters.
+
+With ARG being positive number N, repeat that many times.
+
+With ARG being Negative number -N, repeat that many times in
+forward direction.
+
+See `sp-backward-symbol' for what constitutes a symbol."
+  (interactive "p")
+  (interactive)
```
Fuco1: This line duplicates the one above ;)

smartparens.el
```diff
@@ -8626,6 +8626,31 @@ backward direction."
   (interactive "p")
   (sp-backward-kill-symbol arg t))
 
+(defun sp-backward-delete-symbol (&optional arg word)
+  "Delete a symbol backward, skipping over any intervening delimiters.
+
```
Fuco1: Maybe we should add a clarifying line here explaining that *deleted* text does not go to the clipboard or kill ring.

Fuco1:
Thanks!

Issue #705, opened 2017-02-07
Better integration with hungry-delete?


therockmandolinist:
It seems sp-backward-delete-char clashes with/blocks hungry delete. Is there a way to integrate a fix into smartparens?

There was a workaround posted [on stackexchange](https://emacs.stackexchange.com/questions/29476/always-remove-brackets-by-pairs-fix-hungry-delete-mode-and-smartparens-strict/29484#29484) which may be relevant to this, though it doesn't seem to work very well for me (only works in the scratch buffer?)

Fuco1:
We could probably wrap it and do what we do with subword mode.

Fuco1:
Closed in #764.

PR #764, opened 2017-06-29
Handle hungry-delete-mode.

therockmandolinist:
This would address #705. Not really sure if this is the right way to go about this, but i think it wouldn't interfere with any smartparens operations, only setting the default to `hungry-delete-*` if hungry-delete-mode is on. 

Minimal testing seems to confirm this, though I'm sure I've only covered my biased subset of use cases.


Fuco1:
Seems fine to me, thanks!

PR #767, opened 2017-07-02
Don't affect delete-selection-mode when SP is off

### LINKED ISSUE #763: smartparens interferes with delete-selection-mode when smartparens has been turned off
## Expected behavior

In a buffer where smartparens-mode was on *but is now off*, and where delete-selection-mode is on, inserting any character while region is active always deletes the text in region before inserting the typed character.

## Actual behavior

Region becomes inactive, character is inserted at point.  Selection is not deleted as desired.

1. Start a fresh Emacs (e.g. `mkdir /tmp/emacs && HOME=/tmp/emacs emacs`)
1. Install latest smartparens from MELPA
1. `(require 'smartparens-config)`
1. Go to any Emacs Lisp buffer, such as `*scratch*`
1. `M-x delete-selection-mode RET` to turn on delete-selection-mode
1. `M-x smartparens-mode RET` *twice* to turn smartparens on and back off in the buffer
1. Type "foo"
1. `M-SPC M-b` to select the "foo" you just typed
1. Type "("

I expected "foo" to be deleted and replaced by "(", but instead I now have no active region and the buffer contains "(foo".

- `smartparens` version: 20170619.149 (from MELPA as of a few minutes ago)
- Active major-mode: lisp-interaction-mode
- Emacs version (`M-x emacs-version`): GNU Emacs 25.2.1 (x86_64-apple-darwin15.6.0, NS appkit-1404.47 Version 10.11.6 (Build 15G1421)) of 2017-04-24
- Spacemacs/Evil/Other starterkit (specify which)/Vanilla: Vanilla
- OS: OS X 10.11.6

## Comments

I'm guessing the problem here is that `sp-pair-list` remains a set buffer-local variable when smartparens has ever been turned on in the buffer.  The advice on `delete-selection-pre-hook` calls `sp-wrap--can-wrap-p` which consults `sp--get-pair-list-wrap`.  Neither the advice nor either of these functions ever check if smartparens-mode is currently active in the buffer in question, so the advice never calls `delete-selection-pre-hook` because it expects we'll instead wrap the region later on.

Therefore my first guess at a fix for this is that either the advice or `sp-wrap--can-wrap-p` should check if smartparens is even active in the current buffer.

dsedivec:
When `delete-selection-mode` is on, typing a character with an active region will replace the region with the typed character.  However, smartparens advises `delete-selection-pre-hook` to inhibit this behavior when the typed character is part of a pair, as smartparens wants to wrap the region rather than replace it.

However, if you turn smartparens on and then back off in a buffer where `delete-selection-mode` is also on, this advice would prevent `delete-selection-pre-hook` from replacing the selection even though smartparens is not on, and therefore the region would be neither deleted nor wrapped.  This commit makes it so that the smartparens advice never affects `delete-selection-pre-hook` unless smartparens-mode is turned on.

Fixes #763.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -1393,7 +1393,7 @@ kill \"subwords\" when `subword-mode' is active."
 
 (defadvice delete-selection-pre-hook (around fix-sp-wrap activate)
   "Fix `sp-wrap' in `delete-selection-mode'."
-  (unless (sp-wrap--can-wrap-p)
+  (unless (and smartparens-mode (sp-wrap--can-wrap-p))
     ad-do-it))
 
 
--- a/test/smartparens-wrapping-test.el
+++ b/test/smartparens-wrapping-test.el
@@ -87,6 +87,32 @@
     (sp-test-wrapping "|aM" "[[" "[[|a]]")
     (sp-test-wrapping "Ma|" "[[" "[[a]]|")))
 
+(ert-deftest sp-test-wrap-in-delete-selection-mode nil
+  (sp-test-with-temp-elisp-buffer "|fooM"
+    (delete-selection-mode 1)
+    ;; Inserting a character that pairs should wrap instead of
+    ;; replacing the selection.
+    (execute-kbd-macro "(")
+    (sp-buffer-equals "(|fooM)")))
+
+(ert-deftest sp-test-delete-selection-mode-still-works nil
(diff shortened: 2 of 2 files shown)


Fuco1:
Amazing patch, with tests and all <3  Thanks!

PR #769, opened 2017-07-05
Fix apostrophe in ESS comments

### LINKED ISSUE #745: extra ' in ess comments
## Expected behavior
With `|` representing point in R-mode (from ESS):

```R
## this is a test|
```

Pressing `'` should then only insert one `'` (it does this in emacs-lisp-mode, for example).

## Actual behavior
It inserts 2 `'`

See above

- `smartparens` version: 20170430.844
- Active major-mode: R-mode
- Emacs version (`M-x emacs-version`): 25.2.1
- OS: linux

jabranham:
I'm not super familiar with smartparens, but this seems to fix #745 for me.

### DIFF
--- a/smartparens-ess.el
+++ b/smartparens-ess.el
@@ -161,7 +161,7 @@ ID, ACTION, CONTEXT."
                  :pre-handlers '(sp-ess-pre-handler)
                  :post-handlers '((sp-ess-open-sexp-indent "M-j")))
   (sp-local-pair "'" nil
-                 :unless '(sp-ess-roxy-str-p)))
+                 :unless '(sp-ess-roxy-str-p sp-in-comment-p sp-in-string-quotes-p)))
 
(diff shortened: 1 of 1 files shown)


Fuco1:
Hi, thanks for taking the time to make the patch.  I'm going to merge it because it is better than no solution, but think about the following questions, maybe we can come up with something better:

If the problem is with the insertion of two `'` when you are trying to type a contraction maybe we can use the predicate `sp-point-after-word-p` which is non-nil `here|` but not `here |` where we maybe want to insert a pair.

Is the `''` quote pair useful at all in R?  Does it mean char or string? Maybe we could disable it completely if it's useless.

jabranham:
Thanks, sp-point-after-word-p sounds promising, I'll look into it later. 

Yes, `''` starts characters in R, so we can't just disable it.

PR #777, opened 2017-07-22
Fix handling of commas during slurp and barf in elixir-mode

whysoserious:
Current configuration of smartparens for `elixir-mode` has problems with handling commas (`,`) during slurping and barfing. 
Example: 

I have stolen the solution from `scala-mode` which had the similar problem:

### DIFF
--- a/smartparens-elixir.el
+++ b/smartparens-elixir.el
@@ -27,6 +27,9 @@
 
 (require 'smartparens)
 
+(--each '(elixir-mode)
+  (add-to-list 'sp-sexp-suffix (list it 'regexp "")))
+
 (defun sp-elixir-def-p (id)
   "Return non-nil if the \"do\" keyword is part of definition.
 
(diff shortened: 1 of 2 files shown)


Fuco1:
I'm thinking it would be valuable to simply make this rule the default and add the suffix handling to the modes which require it (C-like in some constructs and text based modes).  It seems, contrary to the original assumption, that the case here is the standard and not the exception.

It would be cool if you could add a test case so that when I refactor this later things wouldn't break.  Do you know about Elisp testing?  We have an article on the wiki but it's more about the conventions we use rather than the setup.  If you can't write the tests that's fine though.

whysoserious:
> I'm thinking it would be valuable to simply make this rule the default

Absolutely, it seems to be the default in most of languages supported by smartparens.

> Do you know about Elisp testing?

I did not, till now ;) 

I added missing tests. Thanks for a quick response!

Fuco1:
Great!  I'm going to merge this as it solves the problem until we move forward with the refactoring.

Thanks!

Issue #346, opened 2014-07-16
Request: show-smartparens and offscreen delimiter


elemakil:
On the [EmacsWiki](http://www.emacswiki.org/emacs/ShowParenMode#toc1) there is a nice `advice` to the builtin `show-paren-function`, which will print the matching delimiter in the echo area if it is offscreen. 

I'd like to put in a request for this feature for the `show-smartparens-mode`. I have adapted the solution given in the EmacsWiki article for `smartparens` (see below code), however, this solution is limited to the _traditional_ paren pairs and also I'd assume that there is a cleverer solution than this.

``` elisp
;; show matching parens on steroids
(defadvice sp-show--pair-function
  (after sp-show--pair-function-offscreen activate)
  "If the matching paren is offscreen, show the matching line in the
        echo area. Has no effect if the character before point is not of
        the syntax class ')'."
  (interactive)
  (if (not (minibuffer-prompt))
      (let ((matching-text nil))
        ;; Only call `blink-matching-open' if the character before point
        ;; is a close parentheses type character. Otherwise, there's not
        ;; really any point, and `blink-matching-open' would just echo
        ;; "Mismatched parentheses", which gets really annoying.
        ;; (if (char-equal (char-syntax (char-before (point))) ?\))
        ;;     (setq matching-text (blink-matching-open)))
        ;; (if (not (null matching-text))
        ;;     (message matching-text)))))
        (let* ((cb (char-before (point)))
               (matching-text (and cb
                                   (char-equal (char-syntax cb) ?\) )
                                   (blink-matching-open))))
          (when matching-text (message matching-text))))))
```

rgemulla:
Another option below. This code directly uses the matching-pair overlay of smartparen (and thus avoids expensive recomputation of the pair or the use of another library).

```lisp
  ;; also show the text of matching paren in echo area
  (require 'subr-x)
  (with-no-warnings
    (make-variable-buffer-local 'rg/sp-previous-overlay)
    (make-variable-buffer-local 'rg/sp-previous-point))
  (defun rg/sp-show-match-in-echo-area ()
    "If the point is on an opening or closing paren and
show-smartparens-mode is active, show the line of the matching
paren in the echo area if it's not visible in the window."
    (interactive)
    (when show-smartparens-mode
      (let* ((start (car sp-show-pair-overlays))
             (end (cdr sp-show-pair-overlays))
             (message-log-max)) ;; don't log in messages
        (if (and start end)
[...]

PR #787, opened 2017-08-31
Implemented request: show-smartparens and offscreen delimiter #346

rgemulla:


Fuco1:
You've opened this PR from your master branch which means it also contains the unrelated latex commit.  Do you know how to put the "show paren" patch on a branch?  It would be great if we could review both separately.

```
git checkout -b show-pair-patch origin/master
git cherry-pick f440c462
```

should do the job.

PR #788, opened 2017-09-03
Implemented request: show-smartparens and offscreen delimiter #346

rgemulla:


Fuco1:
This looks great!

Do you think you could also add some tests? (I'm not really sure how to test this though :D).  If you don't know elisp testing I'll try to do it myself and then we merge it all together.

rgemulla:
I don't know how to test this either. I am running this patch for a while without problems, but that may not mean much.

Anyway, I'd greatly appreciate if you went ahead and add tests (in case you come up with some ;)).

Fuco1:
Yea, I've tried it and it seems to work, but the tests are more to ensure we won't break it later.

I think we could test this with `noflet` and redefinition of `message` and then check the output of that.  I'll see what I can do.

rgemulla:
Note that I added two more commits to this branch. The reason is that when Emacs asked whether a buffer should be saved before killing it, the matching-paren overlay was shown nevertheless and one couldn't see the question one was supposed to answer. This is hopefully fixed now.

rgemulla:
Any idea why this build failed? The information given seems is not really helpful (clojure-mode dependency missing). Not sure if this is actually related to this pull request.

Fuco1:
Nope, not related.  LGTM, thanks for all the work!

Fuco1:
@rgemulla I'm playing with this a bit, wouldn't it be better not to trim the match? So that we can see the indentation where it matches?  I don't know, I think I would prefer such behaviour.  Maybe I'll add a switch for that, what do you think?

rgemulla:
I had also thought about this, but I felt that one couldn't see the indentation well and some space in the echo area is wasted. So an option is a good idea. We may also think about replacing leading spaces by a character that improves readability, such as ␣.

Depending on the language, even more context than the indentation might be helpful. E.g., when the point is on the closing brace in
```c
int f ()
{
...
}
```
we may want to echo the opening brace + the line(s) before. What's best to show seems to be language and personal-preference dependent.

Fuco1:
Yea, it makes sense to me to add a function callback that the major-mode hook could set up and it will be used to calculate the echoed line instead of the default (we can provide a default callback which would work in case nothing is specified specifically).

We can have two callbacks, one for the "opening" and one for the "closing" paren and pick the one depending on the context.

rgemulla:
I like this idea. Together with a function along the lines of ```buffer-substring-with-overlays```, this should give all the flexibility needed.

PR #789, opened 2017-09-13
In rust-mode, allow typing match branches (=>) in strict mode

antifuchs:
I'm writing a bunch of rust lately, and have run into a problem where I couldn't insert `=>` in strict mode. Soooo, this pull request adds a test & adjusts the pair logic to correctly work when `=>` is inserted in rust.

Hope this is acceptable - it looks like all rust tests are passing with this change (:


Fuco1:
Awesome! Even with tests <3

Thanks!

PR #790, opened 2017-09-16
Add `turn-off-smartparens-strict-mode`

xiongtx:
There's an asymmetry in having `turn-on-smartparens-strict-mode` without having a corresponding `turn-off`.


Fuco1:
:heart:

Issue #784, opened 2017-08-28
Paren matching in LaTeX mode


rgemulla:
In LaTeX-mode, parens that follow double backslashes (newlines) are not matched correctly. For example, in tabular environment, I sometimes use `\\[1cm]` to add a custom spacing between successive rows. 

## Expected behavior
The closing square bracket should be matched to the opening square bracket. In general, smartparens should ignore all sequences of double backslashes preceeding a paren.

## Actual behavior
The closing square bracket is marked as mismatched.

## Steps to reproduce the problem
Create a new buffer, switch to LaTeX-mode, then type `\\[1cm]`.

Fuco1:
I think the problem here is that there is also a pair `\[` for the math environment, and it picks this as the "possible opener" instead of just matching `[`.  It should be possible to configure somehow.

rgemulla:
One option may be to define the opener for `\[` as a regexp along the lines of `[^\](\\)*\[` (and the closing expression likewise). This way, it wouldn't match `\\[` as desired, but it would correctly match `\[` and `\\\[`. Can this be done already? 

Another issue may be that the initial sequence of double backslashes shouldn't actually be a part of the matched paren pair, which may be more involved.

Fuco1:
I think the simplest would be to make the `\[` pair be ignored if there is one more slash in front, so something like your last proposal.  This can be done with an `:unless` condition on the `\[` pair.  ISTR we've done something similar already though...

There is the predicate `sp-latex-point-after-backslash` which seems to do exactly that.  Which means something must've changed in the meantime and we probably didn't have a test for the scenario.

Fuco1:
I've debugged it a bit and the problem is with the global skip function `sp--backslash-skip-match` which automatically ignores any pair prefixed with a `\`.  This needs to be adjusted for latex a bit.

rgemulla:
Thanks, I'll look into this.

rgemulla:
I changed that function plus a modified :unless condition everywhere, see commit adfb37e. This greatly improves things, but there are still open issues. In particular:

**Now works correctly**

Insertion and pair highlighting
```
()
\(\)
\\()
\\\(\)
\\\(\\\)
```

Mismatch highlighting:
```
(\)
\\(\)
```

**Still does not work**

Highlighting only shown when point on opening paren:
```
(\\)
```

Backspace deletion removes too much (should only remove "(", but removes "\(")
```
\\(
```

Prefixes not detected
```
\emph{abc}
```

[...]

PR #792, opened 2017-10-03
Ignore double backslashes (newlines) in Latex (#784)

rgemulla:
Fixes the roughest issues of #784; see the discussion there.


rgemulla:
Here is a summary of what this PR fixes and what's still open. I am not sure where to look for fixing the open issues, but I do think that this PR improves handling LaTeX files already substantially. From my point of view, it can be merged.

**Now works correctly**

Insertion and pair highlighting
```
()
\(\)
\\()
\\\(\)
\\\(\\\)
```

Mismatch highlighting:
```
(\)
\\(\)
```

**Still does not work**

Highlighting only shown when point on opening paren:
```
(\\)
```

Backspace deletion removes too much (should only remove "(", but removes "\\(")
```
\\(
```

Prefixes not detected
```
\emph{abc}
```

Tags don't work at all (neither triggers nor highlighting):
```
\begin{quote}
...
\end{quote}
```

rgemulla:
In my point of view, this PR can be merged. There are still some problems open with LaTeX, but the PR nevertheless fixes many of the problems regarding backslashes.

rgemulla:
Not sure why the checks fail; it does not seem to be related to this PR though.

Fuco1:
Yea, tests are failing again, I think something in cask/evm broke.

I'll try to run this locally.  Some additional tests for the new features would also be nice to add.

PR #794, opened 2017-10-10
Improve performance of delete-selection-pre-hook when no region is active

rgemulla:
This is a small change to the ```delete-selection-pre-hook``` advice to avoid unnecessary computations.

I noticed sluggish responses when moving around some buffers; profiling revealed the advice to ```delete-selection-pre-hook``` to be the culprit. In particular, it repeatedly called ```sp-wrap--can-wrap-p```, which ate a large fraction of CPU time. With this patch, ```sp-wrap--can-wrap-p``` is only called when there actually is an active region. The additional condition is directly taken from ```delete-selection-pre-hook```; it is used there to decide whether to do something. With this small change, the problem of sluggish response times went away.


Fuco1:
Thanks, nice catch!

PR #807, opened 2017-11-06
Fix grammar in unmatched expression error message

tchajed:


Fuco1:
Good catch!  Thanks :heart:

PR #822, opened 2017-12-01
Remap kill-whole-line in strict-mode

moyamo:
This command has the potential to create unmatched brackets if it is not remapped.


Fuco1:
Hm, I had no idea `sp-kill-whole-line` even exists. :D  LGTM, thanks for the patch!

PR #824, opened 2017-12-04
Make sp-kill-whole-line remove the newline like kill-whole-line

moyamo:
Currently `sp-kill-whole-line` essentially does a balanced <kbd>C-a C-k </kbd> (delete contents of line up to newline) however kill-whole-line acts more like <kbd> C-a C-k C-k </kbd> (delete contents of line including newline.

I think the smartparens commands with similar names should try to mimic the builtin emacs commands. However this change will probably break the workflow of people who depended on the <kbd> C-a C-k </kbd> behaviour.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -6679,10 +6679,9 @@ Examples:
   (interactive)
   (beginning-of-line)
   (sp-kill-hybrid-sexp nil)
-  (let ((line-start (save-excursion (beginning-of-line) (point)))
-        (buffer-end (save-excursion (end-of-buffer) (point))))
+  (let ((empty-last-line (save-excursion (beginning-of-line) (eobp))))
     ;; We can't kill the line if it is empty and the last line
-    (when (and (sp-point-in-blank-line) (/= line-start buffer-end))
+    (when (and (sp-point-in-blank-line) (not empty-last-line))
       (kill-whole-line))))
 
 (defun sp--transpose-objects (first second)
--- a/smartparens.el
+++ b/smartparens.el
@@ -6679,8 +6679,11 @@ Examples:
   (interactive)
   (beginning-of-line)
   (sp-kill-hybrid-sexp nil)
-  (when (sp-point-in-blank-line)
-    (kill-whole-line)))
+  (let ((line-start (save-excursion (beginning-of-line) (point)))
+        (buffer-end (save-excursion (end-of-buffer) (point))))
+    ;; We can't kill the line if it is empty and the last line
+    (when (and (sp-point-in-blank-line) (/= line-start buffer-end))
+      (kill-whole-line))))
 
 (defun sp--transpose-objects (first second)
   "Transpose FIRST and SECOND object while preserving the
--- a/smartparens.el
+++ b/smartparens.el
@@ -6679,7 +6679,8 @@ Examples:
   (interactive)
   (beginning-of-line)
   (sp-kill-hybrid-sexp nil)
-  (sp-kill-hybrid-sexp nil))
+  (when (sp-point-in-blank-line)
+    (delete-char 1)))
 
 (defun sp--transpose-objects (first second)
   "Transpose FIRST and SECOND object while preserving the
--- a/smartparens.el
+++ b/smartparens.el
@@ -6678,6 +6678,7 @@ Examples:
     (some |long sexp))  ->    |)"
   (interactive)
   (beginning-of-line)
+  (sp-kill-hybrid-sexp nil)
   (sp-kill-hybrid-sexp nil))
 
 (defun sp--transpose-objects (first second)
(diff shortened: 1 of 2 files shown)


Fuco1:
I think there is a customize variable that handles the "also kill newline" behaviour for `kill-whole-line`.  We could teach our version to respect that.

```
kill-whole-line is a variable defined in ‘simple.el’.
Its value is nil

Documentation:
If non-nil, ‘kill-line’ with no arg at start of line kills the whole line.

You can customize this variable.
```

And I think we even did something with it previously regarding `sp-kill-hybrid-sexp`.

moyamo:
I think that variable is for `kill-line` and not the _function_ `kill-whole-line`. The _function_ `kill-whole-line` will always kill the newline where as `kill-line` will only kill the newline if you are at the start of the line and the _variable_ `kill-whole-line` is set.

Fuco1:
In that case I say we should do what Emacs does.  If it breaks someone's workflow and they complain we can add a switch on our side.

I'm not sure the implementation here is necessary then.  Wouldn't it be enough to call hybrid kill and then simply kill the newline? i.e. call `(delete-char 1)` or something similar?  It would probably be much faster than doing a hybrid search again.  Or is there some situation I'm not seeing?

moyamo:
This should be better, the previous version may have also accidentally killed two hybrid-sexps. This will work provided `sp-kill-hybrid-sexp` cannot result in an empty line with extra whitespace.

Fuco1:
Yea, this looks good.  Do you know how to add tests?  It would be nice to have some.  If not, no worries, I will leave this open and push some myself.

moyamo:
Not really, you could probably write the tests faster than I can figure out how to write them :-)

moyamo:
> This will work provided sp-kill-hybrid-sexp cannot result in an empty line with extra whitespace.

So I have actually found a case where `sp-hybrid-kill-sexp` results in an empty line with whitespace and so the current code doesn't work. If you have

    |   Indented text
        More indented text

and then you run `sp-hybrid-kill-sexp` the result is

        |
        More indented text

then after running (delete-char 1) the result is

        |   More indented text

However deleting the line "Indented text" should result in

       |More indented text

without the extra indentation. Instead of `(delete-char 1)` if we used `kill-whole-line` then this issue should go away.

Fuco1:
Makes sense to me!  I still have to push the tests, I'm a bit overworked right now :/

moyamo:
I fixed the broken test and added a new test to check for the bug I found in the previous comment. Should I write more tests?

test/smartparens-commands-test.el
```diff
@@ -679,10 +679,11 @@ be."
 
 (sp-test-command sp-kill-whole-line
   ((nil
-    ("(progn (some |long sexp))" "|")
+    ("(progn (some |long sexp))\n" "|")
```
Fuco1: What happens if there is no newline at the end here? Can we have both variants as a test?  I think it should definitely not error in case there is no "trailing empty line" in the file.

Fuco1:
Thanks! I completely forgot about this :blush: It looks good, I only have one small question (see the review).  If I stop reacting don't be afraid to bump this and yell at me to get this merged!

test/smartparens-commands-test.el
moyamo: It seems to be working fine except that it's beeping. I fixed it anyway.

smartparens.el
```diff
@@ -6679,8 +6679,11 @@ Examples:
   (interactive)
   (beginning-of-line)
   (sp-kill-hybrid-sexp nil)
-  (when (sp-point-in-blank-line)
-    (kill-whole-line)))
+  (let ((line-start (save-excursion (beginning-of-line) (point)))
+        (buffer-end (save-excursion (end-of-buffer) (point))))
```
Fuco1: this can be summed with `(point-max)`.  I think the whole check can be replaced by `(eobp)` (end-of-buffer-predicate) since when there is a non-empty line left it seems to be fine.

moyamo: Thanks, I was looking for that function under the name `end-of-bufferp`, but I couldn't find it.

Fuco1:
Cool! Thanks for your contribution, awesome!

smartparens.el
```diff
@@ -6678,7 +6678,11 @@ Examples:
     (some |long sexp))  ->    |)"
   (interactive)
   (beginning-of-line)
-  (sp-kill-hybrid-sexp nil))
+  (sp-kill-hybrid-sexp nil)
+  (let ((empty-last-line (save-excursion (beginning-of-line) (eobp))))
+    ;; We can't kill the line if it is empty and the last line
+    (when (and (sp-point-in-blank-line) (not empty-last-line))
+      (kill-whole-line))))
```
Fuco1: This should instead use `kill-append` to prevent the situation from https://github.com/Fuco1/smartparens/issues/920

PR #836, opened 2018-01-07
Silence byte compiler warnings

jabranham:


smartparens.el
```diff
@@ -9183,7 +9186,7 @@ support custom pairs."
   "Highlight the enclosing pair around point."
   (interactive))
 
-(defun sp-highlight-current-sexp (arg)
+(defun sp-highlight-current-sexp ()
   "Highlight the expression returned by the next command, preserving point position."
   (interactive "P")
```
Fuco1: If you call this function interactively without the argument being declared does it work? I'm not sure why the spec is as it is though :O

Fuco1:
This is sweet.  We should probably add a check on CI to byte-compile all the files and error on warnings.

jabranham:
There are still a few left after this because I didn't understand the code
and didn't want to break anything. On mobile ATM so can't check that
function. Will do later.

jabranham:
I pushed a fix that doesn't break that function. Here are the remaining byte compiler warnings I see:

```
In sp-scala-wrap-with-indented-newlines:
smartparens-scala.el:56:55:Warning: repeated variable _ in lambda-list
Compiling /home/alex/.emacs.d/lib/smartparens/smartparens-text.el...
Compiling /home/alex/.emacs.d/lib/smartparens/smartparens.el...

In toplevel form:
smartparens.el:3013:1:Warning: Unused lexical variable ‘tag-hook’
smartparens.el:7397:1:Warning: Unused lexical variable ‘allowed-open’
smartparens.el:7417:1:Warning: Unused lexical variable ‘allowed-close’
```

Fuco1:
I quite dislike the fact that we can't have multiple `_` arguments... you have to call them `_1` `_2` etc.  I will look at the other three.  I'll push to this branch if that's OK and then we will merge it all together.  What do you think?

jabranham:
Sounds good to me! Feel free to modify whatever you want.

jabranham:
I looked into how to add a check on travis for byte compiler warnings. Apparently `cask build` will byte compile the project, but it doesn't issue the correct error codes or support setting `byte-compile-error-on-warn` to `t`. 

I suppose we could call it directly with something like `emacs --eval "(setq byte-compile-error-on-warn t)" --batch  -f batch-byte-compile ./*.el` but it needs to be taught about where to find dependencies, and I'm not sure where travis/cask/EVM puts those.

Fuco1:
when you call `cask exec emacs ...` it will set up the load path for you.

jabranham:
Thanks, I don't use cask as you can tell.

I just pushed a commit that should make travis compile the package and error on warnings.

jabranham:
OK, travis is now set up on this branch to compile & error on byte compiler warnings. It's currently failing because there are such warnings :-)

Fuco1:
Thanks for all the great stuff!  I should really try to get this in ASAP so it doesn't go to waste.  Bug me if you think it's taking me too long please! :D

jabranham:
It's quite close to being ready to go. There's just one mystery remaining: why is the CI test succeeding currently?

The emacs byte compile command on smartparens.el exits with a `1` (and it stops there without compiling the other smartparens-*.el files) because of the byte compile warnings/errors, but Travis sees this as a `0`. Could cask be to blame for this somehow?

Also, the first byte compiler warning is marked with the comment `;; TODO: WHAT THE FUCK IS THIS ???11?`

Fuco1:
I...don't even know what that is :D I think it can be removed because the tag feature was also removed, but I'm not really sure at this hour.

About the return value, might be cask, we could try some test to just make emacs die with a specific exit code and see if it eats it up or not.  Cask is a bit unmaintained at this point but I think there is someone to merge fixes if this is a bug indeed.

jabranham:
I've removed it and played around for a few minutes and nothing seems catastrophically broken. I pushed a commit, let's see if it passes the tests on travis.

jabranham:
I'm going to merge this now and open 2 new issues that remain unresolved here:

1. the remaining byte compiler warnings
2. the fact that cask (I think cask anyway) doesn't report the correct exit code when there are byte compile errors.

Fuco1:
Thanks!  I hope it's not an issue with cask because I use it on quite a lot of projects :D And if it always works then my tests are potentially broken everywhere lol.

Issue #671, opened 2016-11-05
Make use of `lexical-binding`


Fuco1:
We should switch smartparens to using `lexical-binding` exclusively

- [x] Analyze where we rely on dynamic binding
- [x] Make the necessary variables into `defvar`s so they stay dynamic
- [x] Enable `lexical-binding`
- [x] Replace `lexical-let` with just regular `let`
- [x] Add arguments to all local functions created with `cl-labels` and similar
- [x] Replace all calls to `run-hook-*` where we used the local symbol dynamically
- [x] Go through special files (major mode support etc) and see which handlers rely on `ok` or `next-thing` and figure out a way how to provide that as an argument

PR #845, opened 2018-01-26
Remove cl package

jabranham:
The need for this was eliminated in #671

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -49,7 +49,6 @@
 
 ;;; Code:
 
-(eval-when-compile (require 'cl)) ; for `lexical-let'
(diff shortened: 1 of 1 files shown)


Fuco1:
Makes sense, thanks!

Fuco1:
@jabranham  You're doing a lot of nice work recently, if you would be interested I can add you to the contributors so you can push small stuff like this directly without me having to "verify" it.  PRs are still nice for bigger chunks of work, if for anything else than to stop people duplicating effort.

jabranham:
Thanks! That would be nice to be able to do. I'll probably run most or everything through a PR just to make sure the tests pass. I can then merge stuff myself if it's small enough like this.

PR #846, opened 2018-01-28
Start testing smartparens with Emacs 26 pretest

jabranham:
The Emacs 26 release is right around the corner, so may as well start
testing with it.


jabranham:
Here's the failure, I'll check it later: 

```
configure: error: You do not seem to have makeinfo >= 4.13, and your
source tree does not seem to have pre-built manuals in the 'info' directory.
Either install a suitable version of makeinfo, or re-run configure
with the '--without-makeinfo' option to build without the manuals. 
Failed! See logs above for error.
```

PR #848, opened 2018-01-29
Use lexical-binding in all files

jabranham:
It was already in use in smartparens.el, this adds it to the mode
specific files too.


jabranham:
This adds a whole lot more byte compile warnings, so there's more work for that, see bug #850

Fuco1:
Hm yea, we probably should've done this right away.  I was afraid some of the code might rely on dynamic binding and did not check.  But if the tests pass... /shrug :D

PR #852, opened 2018-02-02
Do not insert elixir bracketed keywords in strings.

obmarg:
Similar to 1bf46316 this prevents smartparens from inserting a `do`,`end` pair
when you type one of it's keywords inside a string.


Fuco1:
Thanks, this makes sense! I wonder why I didn't think about this case in the first place :blush:

PR #853, opened 2018-02-03
Add support for HTML/SGML based functions in js2-jsx-mode

knu:
js2-jsx-mode is js2-mode with support for XML template literals, so symbol based operations work a bit differently from markup language specific modes.

This tweaks sp-get-thing so that it correctly detects XML tag boundaries and makes most HTML/SGML based functions work properly.

One caveat is that self-closing tags (`<Tag/>`) are not treated as expected, as markup language support in smartparens is based on SGML, not XML.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -607,6 +607,8 @@ Symbol is defined as a chunk of text recognized by
                          web-mode
                          jinja2-mode
                          html-erb-mode
+                         js2-jsx-mode
+                         rjsx-mode
                          )
   "List of HTML modes.")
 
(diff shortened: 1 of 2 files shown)


Fuco1:
Hey, thanks! This looks great.

I like the changes, it would be sweet to have some tests too before we merge.  If you don't know how to write those I will add some or you can read the article on our wiki if you need some pointers about testing in elisp.

I also think we could figure out some way to handle self-closing tags.  If we add a pair `<>` then we could test if there is a closing `/` before the ending and in that case abort the "sgml/xml" parsing routine early.  But that's something outside of the scope of this PR.

knu:
@Fuco1 I've added a minimal test case.

Fuco1:
Great, one last thing.  You will need to add the new mode as a dev dependency to Cask file, or is the mode part of `js2-mode`? (because strangely enough the tests pass :O).

Other than this LGTM!

knu:
Yes, js2-jsx-mode is part of js2-mode, defined in js2-mode.el.  rjsx-mode is a thin extension to js2-jsx-mode, so I believe there's no need for extra tests for rjsx-mode.

Fuco1:
Thanks for your contribution! :heart:

PR #869, opened 2018-03-06
Add crystal to config

brantou:
Sorry, the last pr I did not add crystal to the configuration.


Fuco1:
No worries :)

I've added you as a collaborator. I'm fine with minor fixes like this being pushed directly to master (worst case scenario we can always revert so no big deal).  If you want to do some bigger chunk or want a review feel free to open a PR.

PR #873, opened 2018-03-16
Feature/dwim kill region

Fuco1:


smartparens.el
```diff
@@ -9030,7 +9058,13 @@ If that text is unbalanced, signal an error instead.
 With a prefix argument, skip the balance check."
   (interactive "r")
   (when (or current-prefix-arg
-            (sp-region-ok-p beg end)
+            (-let [(&plist :ok ok :last-good-sexp last-good-sexp) (sp-get-region-info beg end)]
```
Fuco1: We should extract this new block into a separate function as it will be reused multiple times

smartparens.el
```diff
@@ -9054,28 +9088,47 @@ of the point."
         (indent-sexp))
       (sp--back-to-indentation column indentation))))
 
-(cl-defun sp-region-ok-p (start end)
-  "Test if region between START and END is balanced.
+(defun sp-get-region-info (start end)
```
Fuco1: Document this somewhere (check if `region-ok-p` was documented)

PR #874, opened 2018-03-17
ignore arrow function syntax "=>" in javascript-modes

### LINKED ISSUE #872: can not kill-region on arrow functions with rjsx-mode & smartparens-strict-mode
## Expected behavior

should be able to `kill-region` on a javascript arrow function in rjsx-mode with smartparens-strict-mode.

## Actual behavior

`user-error: Can not kill the region: the buffer would end up in an unbalanced state after deleting the active region`

1. create a javascript file and turn on rjsx-mode
2. write a javascript arrow function
```javascript
const test = () => {
  console.log('test')
}
```
3. select whole function and call `kill-region` 

- `smartparens` version: 20180308.1042
- Active `major-mode`: `rjsx-mode`
- Smartparens strict mode: t
- Emacs version (`M-x emacs-version`): GNU Emacs 25.3.1 (x86_64-pc-linux-gnu, GTK+ Version 3.22.26) of 2018-02-09
- Starterkit/Distribution: Vanilla
- OS: gnu/linux

I think  `>` is the problem here since it also needed to be paired in HTML, I wonder if this can be done? 
I really like the idea of smartparens-strict-mode, and want to use it in every mode. I can also press C-k before the function twice to do the same thing, but the muscle memory is really hard to resist(

merrickluo:
close #872 
Do I need to add a test for this?
Also I notice the modes list in `smartparens-javascript.el` is different from `smartparens-config.el`, will this be a problem?

### DIFF
--- a/smartparens-javascript.el
+++ b/smartparens-javascript.el
@@ -46,10 +46,28 @@
 
 (require 'smartparens)
 
+(defvar sp--javascript-modes '(
+                               js-mode
+                               javascript-mode
+                               js2-mode
+                               typescript-mode
+                               )
+  "List of JavaScript modes.")
+
 ;; (|sys).path.append---the dot should not travel with the closing
 ;; paren
-(--each '(js-mode javascript-mode js2-mode typescript-mode)
+(--each sp--javascript-modes
   (add-to-list 'sp-sexp-suffix (list it 'regexp "")))
+
+(defun sp-javascript-skip-match-angle-bracket (_ms _mb me)
+  "Non-nil if we should ignore the bracket as valid delimiter."
+  (save-excursion
+    (goto-char me)
+    (sp--looking-back-p (rx "=>") nil)))
+
+(sp-with-modes sp--javascript-modes
+  (sp-local-pair "<" ">"
+               :skip-match 'sp-javascript-skip-match-angle-bracket))
 
 (provide 'smartparens-javascript)
 ;;; smartparens-javasript.el ends here


Fuco1:
Yea, `smartparens-config.el` should hook to all modes that are used there I think.  `js-mode` and `javascript-mode` is the same thing I belive.

A test would be nice if you know how that works in Elisp.

merrickluo:
I ran into a unexpected behavier when trying to add test, for example, with following code
```javascript
() => {
  console.log('hello world');
}
```
If I mark the first line then call `sp-kill-region`, it will proceed, which leaves
```javascript
  console.log('hello world');
}
```

I suppose provided `skip-match` somehow affect the "{" matching here?

Fuco1:
Is the problem happening only in the test or also in your regular usage?  In the test did you activate the strict mode? (it is not on by default)

merrickluo:
In regular usage, I created a test buffer to try, I also tested with other parens.
`[  ]` and `( )` are also affected, but both single and double quotes  works fine.
examples: 
` => [ a ] ` can become ` a] `
` => ( b ) ` can become ` b)`

I'll try to do more debug when I have the time.

Fuco1:
Thanks, appreciate it!

smartparens-javascript.el
```diff
@@ -46,10 +46,28 @@
...
+
 ;; (|sys).path.append---the dot should not travel with the closing
 ;; paren
-(--each '(js-mode javascript-mode js2-mode typescript-mode)
+(--each sp--javascript-modes
   (add-to-list 'sp-sexp-suffix (list it 'regexp "")))
 
+(defun sp-javascript-skip-match-angle-bracket (_ms _mb me)
+  "Non-nil if we should ignore the bracket as valid delimiter."
+  (save-excursion
```
felipeochoa: You can simplify this check to `(and (string= ms ">") (= (char-before mb) ?=))` to avoid setting up a `save-excursion` and doing all the regex stuff. I think you may even be able to skip the check for ">", but @Fuco1 should confirm.

felipeochoa:
I haven't solved the problem, but I think I've identified it. In `sp-get-thing`, `skip-match` correctly skips the pair handling code, but then there's some code at the end (which I don't grok) that for some reason returns the first symbol inside the sexp. The comment says:

> it can still be that we are looking at a /prefix/ of a
> sexp.  We should skip a symbol forward and check if it
> is a sexp, and then maybe readjust the output.

Unfortunately, the "skip a symbol forward" part moves us past the opening paren/bracket/brace. This looks like it is intentional in `sp-get-symbol` but the docstring for `sp-skip-forward-to-symbol` suggests it  should only be skipping whitespace and comments. Not sure which is correct

Fuco1:
@felipeochoa Ouch, that seems interesting.  I'll try to get to this some day, but can't promise much.  However, I should have much more time in the coming weeks as I used to in the first quarter, so some good things might come!

Fuco1:
Fixed in 04f94b8ce207f9476b39eb397486e87c8507c84e, I modified the code from this PR but there were conflicts so I couldn't merge directly.  Thanks!

PR #876, opened 2018-03-20
Add treatment of single quotes for racket

pmatos:
Nothing should be done for single quotes in racket.

### DIFF
--- a/smartparens-racket.el
+++ b/smartparens-racket.el
@@ -49,6 +49,7 @@
 
 (sp-with-modes '(racket-mode racket-repl-mode)
   (sp-local-pair "`" nil :actions nil)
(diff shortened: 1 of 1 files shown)


Fuco1:
This is already configured in `smartparens-config.el`, is there any reason you are not using that?

pmatos:
is that something I need to explicitly require? Currently I only require smartparens-racket which does not have that set.

Fuco1:
@pmatos Fair enough.  The `-config` file is an "entry point" to the preconfigured setup and it contains the setup for all the lisp languages.  I don't think there's much harm in having this duplicated though for people who only wish to use one language.

pmatos:
OK, Was not aware of that. Thanks for merging.

Issue #881, opened 2018-04-21
Use input-pending-p to short-circuit sp-show--pair-function


aaronjensen:
`sp-show--pair-function` can be relatively slow (150ms in some of my files). When using `show-smartparens-mode` this can be stutters in moving around or typing if it kicks in. 

I've had a lot of success increasing perceived performance by monitoring `(input-pending-p)` in tight loops for things that become irrelevant after the point moves. For example: https://github.com/company-mode/company-mode/pull/720

I tried adding a `(not (input-pending-p))` to the `and` in the `while` of `sp--looking-back` and it fixed the stutter for me. I don't know if that's the most appropriate place to put it and/or if that will have bad effects in other was (it probably will). So I'm curious if there's somewhere off of `sp-show--pair-function` that would be more appropriate to short circuit when there is pending input?

Thanks!

aaronjensen:
Here's a profile fwiw:

```
- timer-event-handler                                            1737  83%
 - apply                                                         1737  83%
  - sp-show--pair-function                                       1596  76%
   - when                                                        1596  76%
    - if                                                         1595  76%
     - progn                                                     1595  76%
      - sp--with-case-sensitive                                  1595  76%
       - if                                                      1595  76%
        - let                                                    1593  76%
         - save-match-data                                       1593  76%
          - let                                                  1593  76%
           - unwind-protect                                      1593  76%
            - progn                                              1593  76%
             - cl-labels                                         1586  76%
              - let                                              1586  76%
               - let*                                            1586  76%
                - cond                                           1583  75%
                 - funcall                                       1583  75%
                  - #<lambda 0x170d4ad7f9184e28>                 1581  75%
                   - let                                         1581  75%
                    - sp-get-thing                               1581  75%
                     - sp--with-case-sensitive                   1581  75%
                      - if                                       1581  75%
                       - cond                                    1581  75%
                        - if                                     1581  75%
                         - if                                    1581  75%
                          - save-excursion                       1581  75%
                           - cond                                1581  75%
                            - cond                               1581  75%
                             - and                               1578  75%
                              - or                               1578  75%
                               - and                             1578  75%
                                - sp--looking-back               1578  75%
                                 - sp--with-case-sensitive               1578  75%
                                  - if                           1578  75%
                                   - let                         1578  75%
                                    - if                         1578  75%
                                     - save-excursion               1578  75%
                                      - save-match-data               1578  75%
                                       - let                     1578  75%
[...]

PR #882, opened 2018-04-22
Break from sp--looking-back early if input is pending

aaronjensen:
As mentioned on #881, this is probably not a good idea. I don't actually know
the effects of allowing sp--looking-back to end early in all cases. For
`sp-show--pair-function` it seems to be fine, but it would probably break other
smartparens operations.

Perhaps a var could be `let` in `sp-show--pair-function` that would allow
`sp--looking-back` to abort in this way?

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -4205,7 +4205,9 @@ sequence, not necessarily the longest possible."
           (save-excursion
             (goto-char from)
             (save-match-data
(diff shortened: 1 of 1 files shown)


aaronjensen:
this also appears to have a pretty significant impact on the time it takes to do a `sp--looking-back` if there is no input. It seems input-pending-p has cost.

PR #900, opened 2018-06-17
disable ' and ` for ocaml and reasonml

Khady:
This is a short version of #886 It only removes pairs that are not pairs in ocaml and reason.


Khady:
Gentle ping :)

Fuco1:
Let's do this for now and if you get back to the original issue we can expand later.  Thanks!

PR #905, opened 2018-07-01
add gerbil (scheme) to sp-lisp-modes

rsbowman:
Gerbil is a scheme on top of Gambit, see https://github.com/vyzo/gerbil.


Fuco1:
Cool, thanks!

PR #907, opened 2018-07-05
Added link to default-configuration

qwerbzuio:
For a user coming to readthedocs, it is not immediately clear where the default configuration can be found (although she might guess that this place is also on Github ;-).


Fuco1:
Thanks!  I really should get back to migrating/updating the docs from github wiki to readthedocs.  It is much better format.

PR #910, opened 2018-07-18
Ability to highlight a pair's content

SamirHafez:
`paren` gives you an overlay that matches pairs and their content if you set its matching style to `expression`. `smartparens` doesn't support this matching style therefore there is no way to highlight the content of a pair.

In this PR I'm proposing a new customizable face `sp-show-pair-match-content-face` that will operate solely on the contents of a pair. 

This has an advantage over the `paren` implementation as `paren` is too rigid in that it doesn't allow you to highlight pairs and their contents independently.

Here's a look at a possible configuration
kapture 2018-07-19 at 10 08 16

And a working scenario (should have used a more visible colour for this example :D)
kapture 2018-07-19 at 10 33 48

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -9223,6 +9223,11 @@ the opening delimiter or before the closing delimiter."
   "The face used to highlight pair overlays."
   :group 'show-smartparens)
 
+(defface sp-show-pair-match-content-face
+  '()
+  "`show-smartparens-mode' face used for a matching pair's content."
+  :group 'show-smartparens)
+
 (defvar sp-show-pair-idle-timer nil)
 
 (defvar sp-show-pair-overlays nil)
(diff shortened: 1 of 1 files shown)


Fuco1:
I like how you implemented this, this way we can have both or neither or just the enclosing pairs.

Can you also edit the documentation https://github.com/Fuco1/smartparens/wiki/Show-smartparens-mode or do you need some special permissions for that? (In which case I can do that).

:+1:

SamirHafez:
@Fuco1 thanks.
Can't edit the wiki though, no permissions, sorry.

Fuco1:
@SamirHafez OK, I'll leave this open until I fix the docs so that I won't forget.  If it takes too much time feel free to poke me!

Otherwise LGTM.

kqr:
Maybe something like this, appended to the bottom of the existing page?

> In addition, smartparens has a third face which is used for the expression between matching pairs. This face is named:
>
>     sp-show-pair-match-content-face
>
> By default, `show-smartparens-mode` does not highlight the contained expression at all. Setting `sp-show-pair-match-content-face` to the same thing as sp-show-pair-match-face emulates setting `show-paren-style` to `expression` when using `show-paren-mode`.
>
> However, it may also be useful to use a different colour for the matching pairs and the contained expression. This is of course possible with `show-smartparens-mode` but not with regular `show-paren-mode`.

Fuco1:
@kqr Why junk, I think it makes perfect sense.  I'll copy the text there right away.  Thanks for the contribution!

Fuco1:
Ok, the docs are live https://github.com/Fuco1/smartparens/wiki/Show-smartparens-mode#customizing-the-looks so I'm merging this.

Thank you both for the work again!

PR #922, opened 2018-08-25
Not show overlay on pair and pair content when there's non-empty region marked

ailisp:
Currently if i set a `sp-show-pair-match-content-face` and mark sexp, the marked region is still using the `sp-show-pair-match-content-face` instead of `region` face. So not put overlay on pair and content to avoid confusion.


smartparens.el
```diff
@@ -9360,17 +9360,18 @@ matching paren in the echo area if not visible on screen."
...
-         (oright (make-overlay (- end clen) end nil t nil)))
-    (setq sp-show-pair-overlays (list oleft omiddle oright))
-    (overlay-put oleft 'face 'sp-show-pair-match-face)
-    (overlay-put omiddle 'face 'sp-show-pair-match-content-face)
-    (overlay-put oright 'face 'sp-show-pair-match-face)
-    (overlay-put oleft 'priority 1000)
-    (overlay-put omiddle 'priority 1000)
-    (overlay-put oright 'priority 1000)
-    (overlay-put oleft 'type 'show-pair)))
+  (unless (use-region-p)
```
Fuco1: I think this check should only apply to the `omiddle` overlay since we still want (do we?) to highlight the parens.  At least that's what I'd prefer.

Fuco1: Maybe we can make it customizable but I'd like to keep backward compatibility where by default the middle is not highlighted and the pairs are even with an active region.

ailisp:
@Fuco1 This version I only hide `omiddle` when there's a non empty region. Actually i was not sure about which choice is better, but your comments definitely make sense to me. And yes, this will be consistent with builtin `show-paren-mode`. Thanks!

Fuco1:
Thanks @ailisp ! Very nice UX improvement.

PR #930, opened 2018-09-28
do not reindent fundamental-mode after kill

kidd:
fundamental-mode looks kinda broken when using a simple `M-DEL`  (`sp-backward-kill-word` in strict mode).

It took me some time to figure this out and I think it's usually what average user might want.

WDYT?

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -606,6 +606,7 @@ Symbol is defined as a chunk of text recognized by
                                              asm-mode
                                              makefile-gmake-mode
                                              haml-mode
+                                             fundamental-mode
                                              )
   "List of modes that should not reindent after kill."
   :type '(repeat symbol)


Fuco1:
@kidd I don't have any strong opinion on this as `fundamental-mode` is rarely used.  Can you provide a sample of text where the behaviour is odd though? Maybe we can find a better solution.

kidd:
Hi,
The missbehavior appears not just in fundamental-mode but in modes where there are no indentation rules for it (where pressing TAB indents the line on level deeper instead of moving the line to a fixed position). I don't know much about indentation in emacs so that's the best description I can give, sorry :(.  EDIT: Looks like it happens when `indent-line-function` is `insert-tab`.

Let's take a buffer in `fundamental-mode` with the contents:
```
foo
bar baz|  ```

if `smartparens-strict-mode` is active, when `sp-backward-kill-word`, it tries to reindent the line, meaning the line gets an extra indentation for no apparent reason.

Now that I look at it this way, maybe we could never reindent if `(eq indent-line-function 'insert-tab)` . This same "bug" happens in many "unknown" file types.

I hope the problem is clearer now.

Fuco1:
Yes that last suggestion seems like a cool heuristic!

kidd:
alright! Closing this for now. Expect a new PR shortly.

PR #936, opened 2018-10-06
Add support to avoid reindenting per indent-line-function

kidd:
Added sp-no-reindent-after-kill-indent-line-functions custom variable
which contains `indent-line-function's that shouldn't be reindented.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -611,6 +611,14 @@ Symbol is defined as a chunk of text recognized by
   :type '(repeat symbol)
   :group 'smartparens)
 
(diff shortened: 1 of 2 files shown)


kidd:
Hey @Fuco1 , here's my PR that should solve #930 (and more).  

I'm leaving the decision of abstracting both `(unless (or ....-modes ...-functions))` to a helper function to you :)

Fuco1:
I think it's fine for now, we can abstract it later if it gets more complicated.  Thanks for also including a test case!

PR #943, opened 2018-12-11
Use while-no-input in show-pair-function.

Alexander-Miller:
The large diff hides a trivial change: I wrapped the whole thing with `while-no-input`. This change makes for a *massive* usability improvement. For example my own theme contains a continuous sexp of nearly 1000 lines. Putting point on its edges would freeze the UI for multiple seconds. With this change there is no freeze at all, only the highlight takes a while to show up.

/edit: CI is all red, but judging from the build history that's not on me.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -9392,58 +9392,59 @@ matching paren in the echo area if not visible on screen."
   (when show-smartparens-mode
     (sp--with-case-sensitive
       (save-match-data
-        (cl-labels ((scan-and-place-overlays
-                     (match &optional back)
-                     ;; we can use `sp-get-thing' here because we *are* at some
-                     ;; pair opening, and so only the tag or the sexp can trigger.
-                     (-if-let (ok (sp-get-thing back))
-                         (sp-get ok
-                           (when (or (and back
-                                          (or (= :end (point))
-                                              (= :beg-in (point))))
-                                     (and (not back)
-                                          (or (= :beg (point))
-                                              (= :end-in (point)))))
-                             (sp-show--pair-create-overlays :beg :end :op-l :cl-l)
-                             (when (and sp-echo-match-when-invisible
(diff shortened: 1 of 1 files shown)


Fuco1:
Hey @Alexander-Miller thanks for the patch.

I've never seen this macro, is it available on older Emacs versions?  I'd like to target at least 25.1.  I myself only moved to 26 a week ago :blush: 

I love some of the documentation involved

> The value of that variable is passed to ‘quit-flag’ and later causes a
**peculiar** kind of quitting.

(emph mine)

Fuco1:
OK, I'm going to do an opportunistic merge and trust you that it works :D  We've had issues with this lag before so if it works great, if not we can simply revert later.

Thanks for taking the time!

aaronjensen:
If it doesn't cause any issues to have that function return `t` then yea this should work. I believe I've run into weirdness with the particulars of how `while-no-input` worked that made it worse than manually checking for pending input, but I don't recall them off the top of my head. Seems worth a shot!

aaronjensen:
Sounds good to me. Technically, it's probably more invasive as it requires the code to be exitable at any point, but that's hopefully fine. I don't know how much global state smartparens uses. I'll report any issues I see, but I'd guess they'd be pretty obvious. So if no new error reports come in it's probably A-OK!

sorawee:
I believe there's a bug in this patch. 

1. In Spacemacs (which uses `smartparens`), occasionally a keypress will be ignored, and the message "Quit" will show up in the minibuffer (and also the `*Messages*` buffer)
2. Before this change, this problem doesn't exist.
3. When I blacklist `smartparens` from Spacemacs, the problem also disappears.

Fuco1:
@sorawee Thanks, I copied it into a new issue #946 let's keep the discussion there!

PR #951, opened 2019-01-11
Fix docstring of sp-backward-sexp

whxvd:


Fuco1:
Thanks, good catch!  Copy-paste, the eternal enemy of programmers :blush:

Issue #946, opened 2018-12-30
Keypress ignored and "Quit" printed in minibuffer


Fuco1:
I believe there's a bug in this patch.

1. In Spacemacs (which uses smartparens), occasionally a keypress will be ignored, and the message "Quit" will show up in the minibuffer (and also the *Messages* buffer)
2. Before this change, this problem doesn't exist.
3. When I blacklist smartparens from Spacemacs, the problem also disappears.

(copied from #943)

sorawee:
@arvinpan confirms that disabling `smartparens` makes the bug disappear here: https://github.com/syl20bnr/spacemacs/issues/11736#issuecomment-450636451

Lokeh:
I can confirm that after recently updating to the latest in spacemacs `develop` branch updating all of my packages, this issue began to appear: many `Quit` messages showing up in `*Messages*` buffer and dropping keystrokes when editing e.g. Clojure files.

Excluding `smartparens` in my `.spacemacs` seems to have fixed this issue for now.

aaronjensen:
@fuco1 worth trying some version of Fuco1/smartparens#882 now instead?

@Lokeh or @sorawee could you try keeping `smartparens` enabled and removing the `while-no-input` from `sp-show--pair-function` and see if it helps?

sorawee:
I rolled back the package to `smartparens-20181028.1005` (which is prior #943) and as I said, there's no problem with this version. Since #943 simply adds `while-no-input`, I think that means removing `while-no-input` helps.

TimoFreiberg:
Can't reproduce (neither by writing some clojure code nor by performing the steps outlined in https://github.com/syl20bnr/spacemacs/issues/11736#issuecomment-450544978)

Maybe the emacs version/OS makes a difference?
My setup:

Arch Linux
GNU Emacs 26.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.30) of 2018-07-05
Spacemacs on develop

sorawee:
Good idea. I should have included this. My setup:

MacOS
GNU Emacs 26.1
Spacemacs on develop

aaronjensen:
I haven't seen this and I have similar specs:

macOS 10.14.2
GNU Emacs 27.0.50 (build 1, x86_64-apple-darwin18.2.0, NS appkit-1671.10 Version 10.14.1 (Build 18B75)) of 2018-12-04
Spacemacs on develop

[...]

PR #958, opened 2019-01-15
Use while-no-input from Emacs 27

aaronjensen:
This should speed up responsiveness without the Quit bug. See #943 and #946.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -9408,42 +9408,44 @@ support custom pairs."
             (sp-show--pair-enc-function ok)))
       (execute-kbd-macro cmd))))
 
-;; Ripped from Emacs 27.0 subr.el.
-;; See Github Issue#946 and Emacs bug#31692.
-(defmacro sp--while-no-input (&rest body)
-  "Execute BODY only as long as there's no pending input.
+(defalias 'sp--while-no-input 'while-no-input)
+(when (version< emacs-version "27")
+  ;; Ripped from Emacs 27.0 subr.el.
+  ;; See Github Issue#946 and Emacs bug#31692.
+  (defmacro sp--while-no-input (&rest body)
+    "Execute BODY only as long as there's no pending input.
 If input arrives, that ends the execution of BODY,
 and `while-no-input' returns t.  Quitting makes it return nil.
 If BODY finishes, `while-no-input' returns whatever value BODY produced."
-  (declare (debug t) (indent 0))
-  (let ((catch-sym (make-symbol "input")))
-    `(with-local-quit
-       (catch ',catch-sym
-	 (let ((throw-on-input ',catch-sym)
-               val)
-           (setq val (or (input-pending-p)
-	                 (progn ,@body)))
-           (cond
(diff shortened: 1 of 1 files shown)


aaronjensen:
No, it includes the bug fix for ‘while-no-input’ from emacs 27 which should fix the quit error.

aaronjensen:
@duianto could you give this patch a shot on Emacs 26 please?

duianto:
It seems to be working as expected.

There are no quit messages or selection cancellations, and there's no delay when the cursor is moved over the opening parenthesis of the function:
```elisp
(defun create-spacemacs-theme (variant theme-name)
```
in:
https://github.com/nashamri/spacemacs-theme/blob/4dde3005613f484d250ea796185b0dddbc094a86/spacemacs-common.el#L107

The local Spacemacs location is in:
c:/Users/username/.emacs.d/core/libs/spacemacs-theme/spacemacs-common.el:107 

The delay does occur with the current version of Smartparens (without this patch).

This delay seems to have been the original reason for adding the previous (now reverted): `while-no-input`
Use while-no-input in show-pair-function. #943

#### System Info :computer:

The patch was applied to: smartparens-20190112.1318

- OS: windows-nt
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 990290e24)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(autohotkey colors emacs-lisp git helm html ibuffer markdown multiple-cursors org spell-checking treemacs version-control)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG SOUND NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS THREADS LCMS2

duianto:
This might be the thread where the function was introduced, at least it's discussed there.
https://lists.gnu.org/archive/html/bug-gnu-emacs/2018-06/msg00340.html

aaronjensen:
@Fuco1 This branch has been rebased and should be ready to merge. Emacs 27 is released now and I've confirmed that this is the released version of `while-no-input`

ThibautVerron:
Aha, I was missing this bit of history. I will port it in my local copy and give it a test.

Maybe to be more future-proof, we would want to just have the macro expand `while-no-input` if `emacs-version` is at least 27?

aaronjensen:
@ThibautVerron sure, updated.

ThibautVerron:
Cool. :)

It's really nitpicking, but is there a good reason for putting the test inside the macro, as opposed to having two different defmacro's depending on the version?

aaronjensen:
@ThibautVerron good idea, I hadn't thought of that. I simply aliased it instead of declaring a macro for Emacs 27+

Fuco1:
I'm going to pull this locally and use it for a day, if it works fine on my prehistoric E26.1 I think it's safe to go ahead.

Fuco1:
Haven't yound any problems so I'm merging this.  Thanks for the work everybody, good job!

PR #969, opened 2019-03-25
Add customizable behavior allowing backward-delete of blank sexps

timor:
This allows deleting empty sexps not only when they are completely empty (closing delimiter follows immediately after opening delimiter), but also if they only consist of white-space.

This is useful for modes where white-space is mandatory after/before delimiters, to get the same behavior for deleting empty sexps.

This also changes the return value of `sp-point-in-blank-sexp` to
return the beginning and ending positions of the enclosing regexp.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -1485,6 +1485,12 @@ kill \"subwords\" when `subword-mode' is active."
   :type 'boolean
   :group 'smartparens)
 
+(defcustom sp-delete-blank-sexps nil
+  "If non-nil, allow deletion of enclosing pair when it only
+  contains whitespace."
+  :type 'boolean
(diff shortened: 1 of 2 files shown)


Fuco1:
Should we also want this in the situation where `(    |)` and we hit `C-d`?  That is the behaviour in strict mode that works with `(|)`.

timor:
> Should we also want this in the situation where `( |)` and we hit `C-d`? That is the behaviour in strict mode that works with `(|)`.

I thought about that, but since I never had that use case, I did not implement it. Should be straightforward, though.  Shall I add it?

Fuco1:
I reworded some docstrings a bit and added a test.  Thank you!

PR #975, opened 2019-04-29
set :end correctly when no next sexp in sp-get-hybrid-sexp

edam:
This seems to work for me.

The logic is that when `sp-forward-sexp` (in `sp--get-hybrid-sexp-end`) returns a next sexp whose `:beg` is before point, it must be because there are no more and it returned the parent sexp. And so, when `:beg` < point, either the end of the line or the end of the parent sexp, whichever is sooner.


edam:
Fixes 974.

edam:
Tests now pass as well as they do without this patch! (i.e., diff of ert-runner is identical, with and without these changes).

PR #976, opened 2019-05-07
fix: avoid insert slash pair in org-mode checkbox.

tangxinfa:


tangxinfa:
Without this modification, when we type "/" in "[]" it will become "[//]",
actually we just want a org-mode checkbox "[/]".

Fuco1:
Yes! :D Thank you this was annoying me for so long.

PR #977, opened 2019-05-19
Add `sp-show-pair-match-priority` option

### LINKED ISSUE #801: Need show-paren-priority to to make the region face take priority
image

Currently smartparens uses hard coded 1000 as priority.

valff:
Fixes #801.


Fuco1:
Excellent!

PR #981, opened 2019-06-22
Integration of smartparens with c/cpp-mode

conao3:
I noticed that even though C is a major language, there is no specific configuration file.

For example, when I'm using `smartparens-strict-mode`, the `c-electric-backspace` is not remapped, which clears the closing parentheses despite `smartparens-strict-mode`.

I looked at below site and added the minimum necessary settings.
Of course, not all the pairs needed to edit C's files have been added, but having the files will make it easier for someone else to edit them.

I wrote `smartparens-c` based on `smartparens-clojure`, but if it violates the policy, please let me know. I'll modify.

ref: https://tuhdo.github.io/c-ide.html#orgheadline56

Or there may be another reason for not having a C specific configuration file. I am sorry in that case. Please close it.

### DIFF
--- a/smartparens-c.el
+++ b/smartparens-c.el
@@ -1,0 +1,52 @@
+;;; smartparens-c.el --- Additional configuration for C/C++ mode.  -*- lexical-binding: t; -*-
+;;
+;; Author: Naoya Yamashita <conao3@gmail.com>
+;; Maintainer: Matus Goljer <matus.goljer@gmail.com>
+;; Created: 23 June 2019
+;; Keywords: abbrev convenience editing
+;; URL: https://github.com/Fuco1/smartparens
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License:
+;;
+;; This file is part of Smartparens.
+;;
+;; Smartparens is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+;;
+;; Smartparens is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


conao3:
I mistake `Created` date.

Fuco1:
There is no other reason than me not being a C programmer and nobody else having written it yet.  So your contribution is very welcome.

There is quite a lot of open issues about C/C++ so maybe this also fixes some of them.

conao3:
Me, too. However, since I use `strict-mode` globally, I was surprised when `smartparens` removed the closing parentheses and created disproportionate parentheses when editing C/Cpp.
I don't write C on a daily basis either, so the main purpose of this PR is to reduce hurdles for further contributors.

conao3:
any concerns?

smartparens-c.el
```diff
@@ -0,0 +1,52 @@
...
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with Smartparens.  If not, see <http://www.gnu.org/licenses/>.
+;;
+;;; Commentary:
+;;
+;; This file provides some additional configuration for Clojure mode.  To use
```
Fuco1: it says Clojure here

Fuco1:
Just one typo I spotted.  Otherwise LGTM.  Sorry for the slowish response time.

conao3:
I fix that. Thanks, for your quick reply!

Fuco1:
Thanks!

conao3:
OK, thanks 🎉

PR #995, opened 2019-11-22
fix linter warnings and some headers

conao3:
Hi! I found many linter warnings so I try to solve them.
But many warnings remained.  I fix easy warnings only.

## Warnings

``` emacs-lisp
Lisp symbol ‘regexp-opt’ should appear in quotes (emacs-lisp-checkdoc)
Lisp symbol ‘regexp-quote’ should appear in quotes (emacs-lisp-checkdoc)
Lisp symbol ‘text-mode’ should appear in quotes (emacs-lisp-checkdoc)
Name lisp should appear capitalized as Lisp (emacs-lisp-checkdoc)
```

## Headers

Some headers wrote Created date as <year>-<mon>-<day> format. Follow smartparens.el.

Fix header comment to follow smartparens.el

Remove unneeded version/keyword/url comment header (should refer main smartparens.el).

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -3331,8 +3331,8 @@ extra boundary conditions depending on parens."
     (sp--wrap-regexp (regexp-opt strings) start end)))
 
 (defun sp--strict-regexp-opt (strings &optional ignored)
-  "Like regexp-opt, but with extra boundary conditions to ensure
-that the strings are not matched in-symbol."
+  "Like `regexp-opt', but with extra boundary conditions to ensure
+that the STRINGS are not matched in-symbol."
   (if strings
       (with-syntax-table
           ;; HACK: this is a terrible hack to make ' be treated as a
@@ -3351,7 +3351,7 @@ that the strings are not matched in-symbol."
     "^\\<$"))
 
 (defun sp--strict-regexp-quote (string)
-  "Like regexp-quote, but make sure that the string is not
+  "Like `regexp-quote', but make sure that the STRING is not
 matched in-symbol."
   (sp--wrap-regexp (regexp-quote string)
                    (string-match-p "\\`\\<" string)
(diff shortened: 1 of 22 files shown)


smartparens.el
```diff
@@ -9582,16 +9584,14 @@ has a pair definition.  If so, we insert the closing pair."
...
+  '(sp-wrap-point
     sp-wrap-mark
     sp-last-wrapped-region
     sp-pair-overlay-list
     sp-wrap-overlays
     sp-wrap-tag-overlays
     sp-last-operation
-    sp-previous-point
-    )
+    sp-previous-point)
```
Fuco1: The paren here is intentionally on next line so diffs are nicer: if one thing is added the diff is "+1" instead of "-1/+2".

conao3: Revert this change.

conao3:
I reverted your mentioned diff. There are any other concerns?

Fuco1:
I'll need to do another review, it's a huuuge change set.  Thanks for your patience!

conao3:
Thanks reply! It's just a ping. I thank you very much for maintaining many packages.

conao3:
Any concerns?

Hi-Angel:
The list of warnings that is in the PR description should go instead to the commit body. Because once it's merged, it's highly unlikely anybody's good read the PR, but people do read git-log.

----------------

With that said, @Fuco1 ping? I also noted that smartparens produces various warnings with native compilation, I think this PR would be a good start *(even though I presume it doesn't fix byte-compilation warnings)*

Fuco1:
@Hi-Angel The fixes here are 90% from check doc so it only changes documentation strings and not actual code.  It is still worth going over though.

conao3:
Hi, sorry for this outdated PR. I am in a busy these days and can't spend time on this PR, so I will close it for now.
I will declare CC0 for commit in this PR, even though it may not be necessary.

Fuco1:
Don't worry, I am sorry for not looking into it sooner.  It's still valuable and I will definitely go through it (I'm now slowly going over all pull requests over all my projects)

PR #1006, opened 2020-02-16
The package is called `rustic` and not `rustic-mode`

bkchr:
This fixes auto loading of the rust config for `rustic`.


bkchr:
Ping @Fuco1

Fuco1:
Sorry for the delay and thanks for the fix!

PR #1007, opened 2020-02-20
Add configuration for Coq

tchajed:
This is somewhat magical to me but it's copied from Doom Emacs' configuration for OCaml (tuareg-mode) and F# and it seems to work nicely: https://github.com/hlissner/doom-emacs/blob/1ab9f8b99d35a451a06c8278457ddfa58c566186/modules/config/default/config.el#L182.

I also tried `(sp-local-pair "(*" "*)")` but with that if you type `(*` you get `(**))`. It's quite possible that smartparens has this bug currently for OCaml and F#.

### DIFF
--- a/smartparens-ml.el
+++ b/smartparens-ml.el
@@ -55,6 +55,14 @@
 (sp-with-modes '(fsharp-mode)
   (sp-local-pair "(*" "*)" ))
 
+(sp-with-modes '(coq-mode)
+  ;; Disable ` because it is used in implicit generalization
+  (sp-local-pair "`" nil :actions nil)
+  (sp-local-pair "(*" "*)" :actions nil)
+  (sp-local-pair "(*" "*"
+                 :actions '(insert)
+                 :post-handlers '(("| " "SPC") ("|\n[i]*)[d-2]" "RET"))) )
+
 (sp-with-modes '(tuareg-mode)
   ;; Disable ` because it is used in polymorphic variants
   (sp-local-pair "`" nil :actions nil)
--- a/smartparens-config.el
+++ b/smartparens-config.el
@@ -128,6 +128,7 @@ ID, ACTION, CONTEXT."
 (eval-after-load 'text-mode                '(require 'smartparens-text))
 (eval-after-load 'tuareg                   '(require 'smartparens-ml))
 (eval-after-load 'fsharp-mode              '(require 'smartparens-ml))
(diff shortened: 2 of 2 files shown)


Fuco1:
Is `Coq` in any way related to ML?  I used it a couple years ago but as far as I remember it's a completely different thing.  So probably a separate config file would be better? Or does it share some setup.

smartparens-ml.el
```diff
@@ -55,6 +55,14 @@
 (sp-with-modes '(fsharp-mode)
   (sp-local-pair "(*" "*)" ))
 
+(sp-with-modes '(coq-mode)
+  ;; Disable ` because it is used in implicit generalization
+  (sp-local-pair "`" nil :actions nil)
+  (sp-local-pair "(*" "*)" :actions nil)
```
paldepind, line 61: It would be nice to also add:

```
;; Disable ' because it is used in pattern-matching
(sp-local-pair "'" nil :actions nil)
```

smartparens-ml.el
```diff
@@ -55,6 +55,14 @@
...
+  (sp-local-pair "(*" "*"
+                 :actions '(insert)
+                 :post-handlers '(("| " "SPC") ("|\n[i]*)[d-2]" "RET"))) )
```
paldepind, line 64: What does this do?

Fuco1: It's a tiny DSL which can specify certain actions to be performed after the pair was inserted and another command was triggered.  See https://github.com/Fuco1/smartparens/wiki/Permissions#insertion-specification

Fuco1:
I added some basic config inspired by this patch.  It handles the comment syntax and backtick and quote.  We can use this as a start for future customization.

Thank you @tchajed and @paldepind for the code and review.

LukeXuan:
With the update, I'm getting the `(**))` sequence after typing `(*`, as reported initially by Tej. Is this the intended behavior or I'm missing something?

tchajed:
I'm also having the `(**))` issue now. This is definitely not the desired behavior, and it isn't what happens in OCaml (tuareg-mode).

I have this in my config now which fixes it:

```elisp
    (sp-with-modes '(coq-mode)
      ;;(sp-local-pair "(*" "*)")
      (sp-local-pair "(*" "*"
                     :actions '(insert)
                     :post-handlers '(("| " "SPC") ("|\n[i]*[d-2]" "RET")))
      )
```

PR #1014, opened 2020-03-07
Fix feature name for C-like modes

wentasah:
The feature that implements both c-mode and c++-mode is called
cc-mode. There are no standalone c-mode and c++-mode features.


Fuco1:
Is this also true on older versions of Emacs?  I don't remember who implemented this feature but I would be a bit surprised if it never worked :D But who knows, it happened before.

wentasah:
According to git, both `c-mode` and `c++-mode` functions reside in `cc-mode.el` since at least 1997.

Fuco1:
@wentasah Thanks for double-checking! Sorry for the late response.

PR #1028, opened 2020-04-21
Clarify sp-pair documentation

wentasah:
The sp-pair function does not only add, but also remove or update pair
definitions. Make it immediately clear at the top of the
documentation. Later parts of the documentation refer to updating of
the definition and without mentioning that sp-pair can also update,
the documentation is confusing.


Fuco1:
Thanks, this is a good clarification!

PR #1047, opened 2020-09-17
Use same function arguments for sp-kill-region as kill-region

dakra:
Some packages like `whole-line-or-region` expect the same
function signature for `kill-region` kind of functions.

Now you can e.g. define

```emacs-lisp
  (defun whole-line-or-region-sp-kill-region (prefix)
    "Call `sp-kill-region' on region or PREFIX whole lines."
    (interactive "p")
    (whole-line-or-region-wrap-region-kill 'sp-kill-region prefix))
```

and call `whole-line-or-region-sp-kill-region` (probably mapped to C-w)
to remove the whole line but only if it keeps the balance.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -9151,19 +9151,22 @@ With a prefix argument, skip the balance check."
     (setq this-command 'delete-region)
     (delete-region beg end)))
 
-(defun sp-kill-region (beg end)
+(defun sp-kill-region (beg end &optional region)
   "Kill the text between point and mark, like `kill-region'.
 
 BEG and END are the bounds of region to be killed.
 
 If that text is unbalanced, signal an error instead.
-With a prefix argument, skip the balance check."
+With a prefix argument, skip the balance check.
+
+If the optional argument REGION is non-nil, the function ignores
+BEG and END, and kills the current region instead."
   (interactive "r")
   (when (or current-prefix-arg
             (sp-region-ok-p beg end)
             (user-error (sp-message :unbalanced-region :return)))
     (setq this-command 'kill-region)
-    (kill-region beg end)))
+    (kill-region beg end region)))
 
 (defun sp-indent-defun (&optional arg)
   "Reindent the current defun.


Fuco1:
I don't know if this is necessary.  The package also does wrap `delete-region` with function `whole-line-or-region-wrap-beg-end` so I think you could simply use that instead of the `...wrap-kill-region` version.  

There seems to be no point to the third argument in `whole-line-or-region-wrap-region-kill` anyhow, since it passes the region boundaries to the function as aguments so the third argument seems to be just a compat layer.

If we can make this work I'd rather not add more complexity.

dakra:
Hmm, ok. I just tried it with 

```emacs-lisp
(defun whole-line-or-region-sp-kill-region (prefix)
  "Call `sp-kill-region' on region or PREFIX whole lines."
  (interactive "p")
  (whole-line-or-region-wrap-beg-end 'sp-kill-region prefix))
```

and this works just as well.
I'm a bit confused why then there even is a different
`..-region-kill` function.
It's the same but it calls 2 macros first,
`whole-line-or-region-filter-with-yank-handler`, and
`whole-line-or-region-preserve-column`.
Which docstrings say "Ensure the current column is kept the same after executing BODY."
and "[...] The binding ensure killed strings have a yank handler attached".

I tried experimenting a bit and using those 2 functions in various
ways (killing and yanking multiple lines etc) but I couldn't find a
difference.

So I think you can close this PR but I'm still curious what I'm
missing and what those macros actually do.

Thanks for your time and this great package.
(PS you should add Github sponsors)

Fuco1:
About the sponsor, it wasn't available in my country but it seems that it now is.  I set it up now but it's pending approval.

dakra:
OK. I'm closing this issue now as

``` emacs-lisp
(defun whole-line-or-region-sp-kill-region (prefix)
  "Call `sp-kill-region' on region or PREFIX whole lines."
  (interactive "*p")
  (whole-line-or-region-wrap-beg-end 'sp-kill-region prefix))
```

seems to work the same.
I'm still not sure what those 2 macros in the ..-kill-region
function do and why there even is another function.

But I couldn't find a difference in my one day test so far.

Thanks for your help.

Fuco1:
fiy: my github sponsor profile is now live :tada:

Issue #1039, opened 2020-06-11
sp-get-string does not recognize string fences


ThibautVerron:
Hi,

## Expected behavior

Be able to treat "generalized strings" (delimited by characters with the syntax property `string-fence`) as paired expression.

## Actual behavior

Such strings are not recognized without ad-hoc configuration.

## Steps to reproduce the problem

In a ruby-mode buffer, with `|%w(test)` (string syntax), `sp-forward-sexp` results in `%w|(test)` instead of `%w(test)|`.

I'm not a ruby user (I noticed the same problem with another major-mode I am maintaining) and I realize by browsing the code that smartparens supplies another `forward-sexp` function for ruby, which does take care of that. But it is not sufficient for all the sp functionality, for example the delimiters of the string are not highlighted. 

It seems that `sp-get-string` only detects standard strings, but that `sp--get-string-or-comment-bounds` does detect those strings with arbitrary delimiters. So it should be possible to handle such strings everywhere, no?

## Environment & version information

- `smartparens` version: 1.11.0 (commit 555626a4)
- Active `major-mode`: `ruby-mode`
- Smartparens strict mode: nil
- Emacs version (`M-x emacs-version`): GNU Emacs 26.3 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.11) of 2019-09-16
- Starterkit/Distribution: Vanilla
- OS: gnu/linux

ThibautVerron:
After looking at it more closely, it looks like `sp-get-string` does recognize such generalized strings, then passes it on to `sp--get-string`, which then... chooses to discard it? Why?

https://github.com/Fuco1/smartparens/blob/555626a43f9bb1985aa9a0eb675f2b88b29702c8/smartparens.el#L5355-L5357

ThibautVerron:
Okay. It seems that I can get the functionality I want by removing this test (and replacing `:op cl` with `:op op`), and changing `sp-get-thing` as follows:

- replacing 

  ```(eq (syntax-class (syntax-after (point))) 7)```

  with 

  ```(memq (syntax-class (syntax-after (point)))  '(7 15))```

- moving the whole block up in the condition list (because otherwise the string delimiters can be recognized as sexp delimiters)

I will try to understand how the testing suite works and make a pull request if things go alright (spoiler: they won't).

ThibautVerron:
[...]

PR #1050, opened 2020-10-08
Feature/generic string

ThibautVerron:
Hi,

This is a pull request for the code related to the generic string delimiters ( #1039 ). You might want to squash the history, the last commits were very messy (sorry about that).

It turns out that this:

> As it is, it might be slightly breaking for ruby users, for example sp-backward-sexp with %w(asd)| will now move the point to before %, not before the opening paren.

is no longer a problem. 

But on the other hand, I don't know how to do it again if we were to decide that it is the wanted behaviour. Something seems to be overriding the detection of the string somewhere in sp-get-expression.

Emacs' default sexp navigation (using syntax tables) correctly navigates to the beginning of the string in the same scenario.

Thanks for your work ! :)


ThibautVerron:
> You might want to squash the history, the last commits were very messy (sorry about that).

Or should I do it and offer to merge a clean branch?

Fuco1:
No problem with the squash, GitHub can do it automatically when merging.  Thank you for the PR, I'll review it shortly.

ThibautVerron:
Ping? :)

PR #1054, opened 2020-11-16
Fix compilation of sp--while-no-input

mnewt:
This is a fix for https://github.com/Fuco1/smartparens/pull/958 when byte-compiling. The top level forms need to be wrapped in `eval-when-compile`.

See https://lists.gnu.org/archive/html/bug-gnu-emacs/2020-11/msg01056.html

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -9408,44 +9408,45 @@ support custom pairs."
             (sp-show--pair-enc-function ok)))
       (execute-kbd-macro cmd))))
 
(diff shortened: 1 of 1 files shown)


thblt:
Is there anything blocking this PR? I can confirm it solves the issue for me, using Borg.

Fuco1:
Nope all is fine, just me being unavailable.  Thank you for the fix!

Issue #1024, opened 2020-04-12
[RFE] Do not add a pair if a word is just after the caret


Hi-Angel:
I'm wondering if authors would welcome contribution to change behavior of inserting a pair so, that if a caret is before a word, a pair would not be created. That is because usually when the caret is just before word, more often than not you want to enclose the following text inside parentheses/braces, so you have to manually remove the pair created.

I have [this config](https://github.com/Hi-Angel/dotfiles/blob/bbd08c6883daed98b9feaad7f86304d332f51e3d/.emacs.d/lisp/sp-sublimelike.el) which does that. I called it "sublime-like" because when I created this rule I was inspired by SublimeText, which has amazingly non-intrusive autopairs. That said, I haven't used ST for many years and modified config at some point, so I can't be sure it still matches their behavior. Still, this config I also used for years and I'm pretty satisfied with this behavior.

Hi-Angel:
Okay, thanks! Right, well, it's just that I remember that back when I crafted up that heuristic, my colleagues were using CodeBlocks IDE, which has no heuristic for autopairs. So pressing a `(` always gets you a `)`. I remember nobody liked that feature, it was causing too many misses which was irritating. So nobody used autopairs in CodeBlocks. Judging from this experience, I am not sure that always inserting a pair makes good defaults.

Anyway, you're the boss, thank you :)

Fuco1:
I wouldn't mind shipping some other configuration stacks, but not enabled by default.  In other words, people can pick and choose what kind of strictness they prefer.  Something like "profiles".

Hi-Angel:
Okay, apparently, [someone else](https://www.reddit.com/r/emacs/comments/kcvimc/how_to_move_inserted_semicolon_after_closing/gfyyip2?utm_source=share&utm_medium=web2x&context=3) *(a link to a reddit comment — my browser for some reason doesn't scroll down to it, but it's highlighted)* is interested in such behavior, so I might send a PR then with this being disabled by default.

> I wouldn't mind shipping some other configuration stacks, but not enabled by default

Okay, so, if there's no oppositions, I think of creating a PR with [this file](https://github.com/Hi-Angel/dotfiles/blob/master/.emacs.d/lisp/sp-sublimelike.el) being added similar to, for example, `smartparens-c.el`. *(well, I'll add a header similar to one in the file and a comment about why it's called "sublime-like")*.

PR #1059, opened 2020-12-16
sp-sublimetext-like.el: provide SublimeText-like config

Hi-Angel:
So, per [discussion here](https://github.com/Fuco1/smartparens/issues/1024#issuecomment-746215289), this is an opt-in configuration for parentheses autoinsertion, that supposed to have much less misses *(i.e. situations where a user has to manually delete the pair just inserted)* than the default "always insert a pair" behavior.

It is called sublimetext-like, because originally I have written this config following what SublimeText does. However that was years ago, so these days I don't know if it still matches. Anyway, AFAIR this editor has amazingly unobtrusive behavior for inserting pairs, so if someone decides to make further changes to the code to achieve even closer behavior, they're welcome.


Fuco1:
Thanks, LGTM!

PR #1063, opened 2020-12-23
smartparens-python.el: auto-insert a colon in function definitions

Hi-Angel:
I'm slowly upstreaming various helpers from my Emacs config. This one I've used for a few years I think. Never was getting any misses *(the context for function definition in python is pretty unambiguous)*, so barring something I might have missed while preparing the code for upstreaming, this should be safe.

### DIFF
--- a/smartparens-python.el
+++ b/smartparens-python.el
@@ -47,18 +47,42 @@
 
 (require 'smartparens)
 
+(defgroup smartparens-python ()
+  "smartparens settings for python-mode"
+  :group 'smartparens)
+
+(defcustom sp-python-insert-colon-in-function-definitions t
+  "Whether to auto-insert a colon when parens insertion is
+triggered while typing a function definition"
+  :group 'smartparens-python
+  :type  'boolean)
+
 ;; Python has no sexp suffices.  This fixes slurping
 ;; (|sys).path.append---the dot should not travel with the closing
 ;; paren
 (--each '(python-mode inferior-python-mode)
   (add-to-list 'sp-sexp-suffix (list it 'regexp "")))
 
+(defun sp-python-maybe-add-colon-python (_id action _context)
+  "Adds a colon after parentheses in a python function definition"
+  ;; here, caret supposed to be in between parens, i.e. (|)
+  (when (and sp-python-insert-colon-in-function-definitions
+             (eq action 'insert)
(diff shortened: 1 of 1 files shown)


Hi-Angel:
Even, wait a second, I figured I can improve it by moving the `(forward-char)` to after successful detection

Hi-Angel:
Done, it should be better now.

Hi-Angel:
**upd**: eliminated an excess `when`

Fuco1:
This is very fortunate as I just started writing more python :D :+1: 

I think it would be nice with a configuration option.  We can enable it by default so people can discover it, but then those who don't like it can disable it with a setting.

We can start a group `smartparens-python` if it doesn't exist yet.

Hi-Angel:
> This is very fortunate as I just started writing more python :D 

^_____^

> I think it would be nice with a configuration option. We can enable it by default so people can discover it, but then those who don't like it can disable it with a setting.
> 
> We can start a group `smartparens-python` if it doesn't exist yet.

There is `smartparens-python`. Is it enough, or do you want me to surround the colon auto-completion specifically with an opt-out option?

Fuco1:
Yea, I ment a `defgroup`, similar to 

```
(defgroup smartparens ()
  "Smartparens minor mode."
  :group 'editing
  :prefix "sp-")
```

so that the options for python are grouped under that (setting `smartparens` group as parent).  The opt-out option would then be placed there.

Hi-Angel:
Done now, please see if it looks okay. I didn't work with groups before, but hopefully I did everything correctly.

Fuco1 (changes requested):
all the functions and variables need to have the `sp-python` prefix.

the group stuff seems fine to me

Hi-Angel:
> all the functions and variables need to have the `sp-python` prefix.

Oh, yeah, right, done. I kinda thought a group limits a visibility scope, but I admit that was silly of me, in Lisp everything is global besides `let` expression.

Fuco1 (approved):

Fuco1:
> I kinda thought a group limits a visibility scope, but I admit that was silly of me

I also thought so, especially since you can also specify the `:prefix` parameter to the group which kind of suggests it's going to auto-prefix or something.  But it's only related to some UI things in customize.

LGTM, thanks for the patch!

PR #1066, opened 2021-01-05
Minor refactoring

sirikid:


smartparens.el
```diff
@@ -8388,23 +8379,20 @@ Examples:
...
   (cond
    ((equal arg '(4))
-    (-when-let (items (sp-get-list-items))
-      (let ((op (sp-get (car items) :op))
-            (cl (sp-get (car items) :cl))
-            (beg (sp-get (car items) :beg))
-            (end (sp-get (car items) :end)))
-        (!cdr items)
-        (setq items (nreverse items))
+    (-when-let [(first-item rest-items) (sp-get-list-items)]
```
Fuco1, line 8382: This binds the first two items, not the first and the rest.  For that the pattern has to be `(first-item . rest-items)`.

Fuco1 (changes requested):

Fuco1:
Thanks for taking the iniciative @sirikid.  Unfortunatelly I can't accept all these changes as we have a soft policy of not using cl-lib.

sirikid:
> we have a soft policy of not using cl-lib.

I'm afraid you are already using it, or is this a rule for new code?

Fuco1:
It's sometimes accepted with new features if they use it extensively (if not we try to avoid it), but refactoring existing code to use cl would not fall under this category.

I'm not sure if this is documented anywhere and it's probably a bit frustrating.  I'll try to add it somewhere to the contribution notes.

sirikid:
I don't like `cl-loop` either, but there is no other way to work with plists.

I could write something like this, but it will take the reader a while to realize that I am actually going through the plist.
```elisp
(defun sp--merge-pairs (old-pair new-pair)
  (let ((it new-pair) prop)
    (while it
      (setq prop (pop it))
      (pop it) ;; skip value
      (sp--merge-prop prop new-pair old-pair)))
  old-pair)

(defun sp--update-pair (new-pair old-pair)
  (let ((it new-pair) prop val)
    (while it
      (setq prop (pop it)
            val (pop it))
      (unless (and (plist-get old-pair prop)
                   (equal '(:add) val))
        (plist-put old-pair prop val))))
  old-pair)
```

sirikid:
@Fuco1 could you take another look? This time I tried using "sp code style". There is still one new cl-case,  but if you are totally against it, I can switch it back to cond.

smartparens.el
```diff
@@ -2154,24 +2153,21 @@ old definition with values from PAIR."
...
-         ((eq prop :post-handlers-cond)
-          (--filter (listp it) (plist-get pair :post-handlers)))
-         ((eq prop :when)
-          (--filter (not (listp it)) (plist-get pair :when)))
-         ((eq prop :when-cond)
-          (-flatten (-concat (--filter (listp it) (plist-get pair :when)))))
-         (t (plist-get pair prop)))
-      pair)))
+    (cl-case prop
+      ((nil) pair)
```
Fuco1, line 2157: Is this matching `nil` or `(nil)`, these are not the same!

smartparens.el
```diff
@@ -756,9 +757,10 @@ after the smartparens indicator in the mode list."
       (progn
         (unless smartparens-mode
           (smartparens-mode 1))
-        (unless (-find-indices (lambda (it) (eq (car it) 'smartparens-strict-mode)) minor-mode-overriding-map-alist)
```
Fuco1, line 759: Isn't this the same as `(assoc 'smartparens-strict-mode minor-mode-overriding-map-alist)`.  I think we're both overcomplicating it.

Fuco1:
@sirikid Thanks, these changes look very good.  I just have some additional questions, see the review.

smartparens.el
sirikid: This is matching `nil`. You can't use short form (single symbol) for `nil` because it's also a list on itself.

smartparens.el
sirikid: Even `assq`.

sirikid: I think `alist-get` would be ideal, but it was added in 25.1
```elisp
;; add 
(setf (alist-get 'smartparens-strict-mode minor-mode-overriding-map-alist)
      smartparens-strict-mode-map)
;; remove
(setf (alist-get 'smartparens-strict-mode minor-mode-overriding-map-alist nil t)
      nil)
````

smartparens.el
Fuco1: Oh, good to know, I'm not a huge `cl-case` user. :+1:

Fuco1:
Thanks for the improvements!

bymoz089:
this merge introduces a bug, see #1077

manuel-uberti:
Could it be a slip because of too much Clojure programming? :)

sirikid:
> Could it be a slip because of too much Clojure programming? :)

This is because dash is inconsistent and has different rules for `-let` and `-when-let`

manuel-uberti:
> > Could it be a slip because of too much Clojure programming? :)
> 
> This is because dash is inconsistent and has different rules for `-let` and `-when-let`

As I said [here](https://github.com/Fuco1/smartparens/issues/1077#issuecomment-779683117), isn't `when-let` already available in Emacs? Does the `dash` version offer a better implementation?

sirikid:
> Does the `dash` version offer a better implementation?

It also offers destructuring.

duianto:
There's a PR pending with a fix: [Fix #1077 #1079](https://github.com/Fuco1/smartparens/pull/1079)
for:
>Eager macro-expansion failure: (wrong-type-argument listp [(first-item . rest-items) (sp-get-list-items)])

duianto:
The PR has been merged, the error message doesn't appear anymore.

PR #1078, opened 2021-02-14
WIP: Use ppss accessors instead indices

sirikid:
Seems like ppss once became a structure and now it has accessors, I think it makes sense to use them instead of indices.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -1829,7 +1829,7 @@ fences.
 If optional argument P is present test this instead of point."
   (ignore-errors
     (save-excursion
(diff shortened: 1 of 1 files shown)


Fuco1:
Ugh, I hope they really don't do this without backward compatibility.

sirikid:
> Ugh, I hope they really don't do this without backward compatibility.

It is fully backward compatible. Something like that should be fine:

```elisp
(defalias 'sp--ppss-string-terminator 
  (if (version< emacs-version "27.1")
      (lambda (ppss) (nth 3 ppss))
    #'ppss-string-terminator))
```

PR #1086, opened 2021-04-01
Backport ppss accessors and more

sirikid:
- Backport appropriate `ppss` accessors, they were added in Emacs 27.1.
- Remove backported `defvar-local` because smartparens already requires Emacs 24.4 and `defvar-local` was added in 24.3.
- Add emacs to the dependencies in the package file.

This PR does not introduce new failing tests.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -82,17 +82,6 @@
 
 ;;; backport for older emacsen
 
-;; introduced in 24.3
-(unless (fboundp 'defvar-local)
-  (defmacro defvar-local (var val &optional docstring)
-    "Define VAR as a buffer-local variable with default value VAL.
-Like `defvar' but additionally marks the variable as being automatically
-buffer-local wherever it is set."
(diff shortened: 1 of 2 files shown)


smartparens.el
```diff
@@ -4344,26 +4349,22 @@ The context at point is considered the reference context."
 POINT defaults to `point'.
 
 If the point is not inside a quoted string, return nil."
-  (setq point (or point (point)))
   (save-excursion
-    (goto-char point)
```
Fuco1, line 4349: I think you can't remove this as later in the code `(point)` is used, which could refer to a different place.

sirikid: I think I can:

> (syntax-ppss &optional POS)
> Parse-Partial-Sexp State at POS, defaulting to point.
> ...
> Point is at POS when this function returns.

sirikid: This is the first paragraph from the syntax-ppss help

PR #1087, opened 2021-04-21
Also detect minibuffer-mode

yangsheng6810:
Recently, on Emacs master branch, a commit 55db25b257 was pushed to fix
Emacs bug [47150](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=47150). This bug was related to an incorrect implementation of
minibuffer-inactive-mode. It used to be the case that after the first
usage of the minibuffer, it stucks in minibuffer-inactive-mode, though
minibuffer-inactive-mode was intended for a completely different
purpose, and the minibuffer on the current frame should NOT be inactive.
The new design is to have two possible major modes for the minibuffer:
minibuffer-mode when it is active, and minibuffer-inactive-mode when it
is inactive.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -785,6 +785,7 @@ after the smartparens indicator in the mode list."
   :group 'smartparens)
 
 (defcustom sp-ignore-modes-list '(
+                                  minibuffer-mode
                                   minibuffer-inactive-mode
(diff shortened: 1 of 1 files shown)


Fuco1:
Thanks!

PR #1099, opened 2021-08-13
Add support for triple quoted string literals in Scala

### LINKED ISSUE #1098: Add support for triple quotes in Scala
## Expected behavior

Smartparens should close [multi-line string literals](https://www.scala-lang.org/files/archive/spec/2.13/01-lexical-syntax.html#string-literals) in Scala.

## Actual behavior

Smartparens doesn't close those.

In scala-mode type in `"""`, note that the result is not balanced

- `smartparens` version: 20210529.1129
- Active `major-mode`: `scala-mode`
- Smartparens strict mode: nil
- Emacs version (`M-x emacs-version`): GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.23, cairo version 1.16.0) of 2021-01-18, modified by Debian
- Starterkit/Distribution: Spacemacs
- OS: gnu/linux

nigredo-tori:
Fixes #1098

This is not very robust: inserting double quotes inside a literal doesn't seem to count towards the closing lexeme, so there's no "typing yourself out of the literal". Still, it should be better than nothing. Any pointers would be appreciated.

### DIFF
--- a/smartparens-scala.el
+++ b/smartparens-scala.el
@@ -76,5 +76,7 @@
                                 ("| " "SPC")
                                 sp-scala-wrap-with-indented-newlines))
 
+(sp-local-pair 'scala-mode "\"\"\"" "\"\"\"")
+
 (provide 'smartparens-scala)
 ;;; smartparens-scala.el ends here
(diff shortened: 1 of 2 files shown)


Fuco1:
I think the situation where you open with `"""` then type stuff and then type `"` three times to "close" is going to be somewhat problematic.  I think in python we tried to do the same thing but gave up and so now it only inserts the closing tripple quote when you input three of them and that's it.  Any other fixes you need to do manually (for example with `C-q "`).

I'll merge this as it is better than not having it.  Thanks!

PR #1102, opened 2021-09-11
Scala wrap indented region correctly

VlachJosef:
Hello, 

thanks for awesome package, here is a fix for wrapping region with `{` in scala.

Without this fix result of `sp-test-scala-curly-wrap-indent` test would be: 
```
"if (true)\n{fo\n  o\n}"
```
Instead of correct
```
"if (true)\n{\n foo\n}"
```


Fuco1:
Diki.  Good catch!

PR #1104, opened 2021-09-30
Make sp-forward-symbol and sp-backward-symbol skip comments.

pierre-rouleau:
The change diff seems bigger that it really is.  The change include:
- A new function: `sp-inside-comment` that returns non-nil when point is inside comment
- Modification of the sp-forward-symbol`: add form inside the cond form to move forward while in comment in the first loop.
- Modification of the `sp-backward-symbol` : add form inside the cond form to move backward while in comment in the first loop.


Fuco1:
Well, this works like that on purpose.  If we want such a feature it would need to be behind a feature flag because this is backward incompatible behaviour change.  Some 3rd party code might depend on this.

pierre-rouleau:
Then perhaps it'd be good to document the current behaviour.  I'll look at the feature flags and see if I can add that under feature flag.

PR #1108, opened 2021-10-02
add Fennel to lisp modes

andreyorst:
[Fennel](https://fennel-lang.org/) is a Lisp like language that compiles to Lua

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -570,6 +570,8 @@ Symbol is defined as a chunk of text recognized by
                            common-lisp-mode
                            emacs-lisp-mode
                            eshell-mode
+                           fennel-mode
+                           fennel-repl-mode
                            geiser-repl-mode
                            gerbil-mode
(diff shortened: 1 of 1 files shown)


andreyorst:
@Fuco1, sorry to disturb you, but can we get this merged? I've recently rewritten fennel-mode to inherit from `prog-mode` instead of `lisp-mode` and because of that Smartparens began to misbehave when inserting quotes.

It's not a big deal, as users just need to add `(add-to-list 'sp-lisp-modes 'fennel-mode t)` before calling the `(require 'smartparens-config)` but I do get occasional questions from Fennel mode users about why quotes are paired after the update.

Fuco1:
Nah, no disturbance at all.  I got a bit flustered with life so my open source work suffers :/

Thanks for the reminder!

andreyorst:
Thanks!

PR #1118, opened 2021-12-21
Add more pairs for OCaml

sangwoo-joh:
Thanks for the great package!
This adds more smart pairs for OCaml languages:

```ocaml
module ... end  (* module *)
sig .... end  (* signature *)
{| ... |}  (* multi-line string *)
[| ... |] (* array *)
```


Fuco1:
Thanks and sorry for the delay!

PR #1121, opened 2022-01-16
Add sp-handle-enter.el

semenInRussia:
Add file `sp-handle-enter.el`. This file handle of <kbd>Enter</kbd> for Smartparens.

By default this able has only `Clang`, but this file add handling of <kbd>Enter</kbd> to all modes (which need to this).

Super Big thank you for this super useful package!!!!!

## List of My Valid Actions
- [ ] Tests
- [x] Documentation in Package
- [x] Valid Naming of Commits
- [x] Valid Naming of Functions/Variables

### All Modes without `Haskell`
```js
function hiFromRussia(arg) {|}
```

Press <kbd>Enter</kbd>:

```js
function hiFromRussia(arg) {
    |
}
```
### For `Haskell`
```haskell
type Locator a = {|}
```

Press <kbd>Enter</kbd>

```haskell
type Locator a = {|
                 }
```

This is standard of formating for `Haskell``

### DIFF
--- a/sp-handle-enter.el
+++ b/sp-handle-enter.el
@@ -1,0 +1,107 @@
+;;; sp-handle-enter.el --- Handle situation when insert pair and press enter  -*- lexical-binding: t; -*-
+;;
+;; Author: Semen Khramtsov <hrams205@gmail.com>
+;; Created: 16 January 2022
+;; Keywords: convenience editing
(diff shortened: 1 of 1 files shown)


semenInRussia:
Sorry for my code!

PR #1125, opened 2022-04-30
Add basic 'swift-mode' support

woolsweater:
- Add pairing for '<' and '>', adapted from the existing code for Rust.
- Add pairing for triple quotes

This is _heavily_ based on the existing Rust support, so I have also credited that code's author in my new file.


Fuco1:
Looks good!  Thanks for also including tests :)

PR #1127, opened 2022-06-22
go-mode configs

wyuenho:
Just adding a bunch of configs I've stolen from other language-specific config files for go-mode


Fuco1:
Looks like a good start config!  Thanks!  Hopefully this will also make it easier for others to contribute more interesting stuff for go.

PR #1137, opened 2022-09-15
Fix `module...end` to `struct..end` for OCaml

sangwoo-joh:
Thanks for the great package.
I added some pairs for OCaml in #1118, but I found my mistake. This fixes `module ... end` to `struct ... end`.
Thank you!


Khady:
(if you wanted to add more, there are a bunch of other pairs in https://github.com/Fuco1/smartparens/pull/886)

Fuco1:
@Khady thanks for the bump :D somehow this slipped my attention :/  @sangwoo-joh Thanks for the fix!

PR #1138, opened 2022-09-23
fix typo

strangepr0gram:
enviroment -> environment


Fuco1:
Thanks for catching these :blush:

PR #1139, opened 2022-11-09
Add pairs for doxygen formulas in C / C++

ax487:
There is a little-known [feature](https://www.doxygen.nl/manual/formulas.html) inside doxygen, which allows users to add inline formulas to the documentation:

```c
  // The distance between \f$(x_1,y_1)\f$ and \f$(x_2,y_2)\f$ is 
  // \f$\sqrt{(x_2-x_1)^2+(y_2-y_1)^2}\f$.
```

This patch adds the corresponding pairs to the `sp-c-modes`.


Fuco1:
Thanks! I didn't know about this either :)

PR #1148, opened 2023-02-24
Fix compiler warnings

sonofjon:


sonofjon:
For reference:

https://emacs.stackexchange.com/questions/73047/emacs-29-docstring-single-quote-escaping-rules-compiler-level-event
https://emacs.stackexchange.com/questions/72062/what-is-in-doc-strings

Fuco1:
LGTM, Thanks!

Issue #1100, opened 2021-08-24
smartparens breaks LaTeX quotation marks mechanism for language styles


vitaminace33:
## Expected behavior
Press `["]` once to get `«.»`, twice for `"."` (or vice-versa if `TeX-quote-after-quote` is non-nil).
#### Notes
- Brackets indicate a key.
- Dot indicates the point or cursor position.
- Double angular brackets are the hypothetical current value of `TeX-quote-open` and `TeX-quote-close` but could be different depending on the applied language style, e.g. `"<` and `">`, respectively.
- Check [AUCTeX's reference manual](https://www.gnu.org/software/auctex/manual/auctex/Quotes.html) for regular (non-electric) behavior.

## Actual behavior
Press `["]` once, depending on the value of `TeX-quote-after-quote` and if `TeX-quote-open` and `TeX-quote-close` are single or double characters, different behavior are present. Worst case scenario is that only part of the quote string is inserted, moreover partly in the previous line and partly in the current, while deleting a previous character. Refer to syl20bnr/spacemacs#15003 for specific "experiments".

## Steps to reproduce the problem
1. On vanilla Emacs, install `auctex` and `smartparens`.
2. Open a .tex file including `\usepackage[german]{babel}` (parsing on load must be activated to successfully load the language style).
3. Write any quote. Again, refer to syl20bnr/spacemacs#15003 for specific "experiments".

I leave here my `.emacs` for reference.
```elisp
(require 'package)
(add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/") t)
(package-initialize)
(require 'smartparens-config)
(setq TeX-parse-self t) ; Enable parse on load.
(setq TeX-auto-save t)  ; Enable parse on save.
(custom-set-variables
 '(package-selected-packages '(smartparens auctex)))
```

## Workaround
A workaround has been documented in syl20bnr/spacemacs#15003 that allows the use `smartparens`.

## Environment & version information
- `smartparens` version: 20210817.1912
- Active `major-mode`: `latex-mode`
- Smartparens strict mode: nil
- Emacs version (`M-x emacs-version`): GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.20, cairo version 1.16.0) of 2020-09-19
- Starterkit/Distribution: Vanilla
- OS: gnu/linux (Ubuntu 20.04)

thobl:
[...]

PR #1151, opened 2023-03-16
Latex fixes

dschrempf:
Hi! This fixes https://github.com/Fuco1/smartparens/issues/1100; well, it disables some features of `smartparens` for latex modes. The PR also includes some other nits and changes. Let me know if I should exclude those.

Finally, this change is a breaking change for people using
```emacs-lisp
(sp-local-pair modes "``" nil ...))) ;; or
(sp-local-pair modes "`" nil ...)))
```

For example, this affects Doom Emacs, which I am using. The corresponding patch for Doom Emacs is attached.
[s.patch](https://github.com/Fuco1/smartparens/files/10995599/s.patch)

### DIFF
--- a/smartparens-latex.el
+++ b/smartparens-latex.el
@@ -179,13 +179,7 @@ ID, ACTION, CONTEXT."
   (sp-local-pair  "\\lvert" "\\rvert"
           :when '(sp-in-math-p)
           :trigger "\\lvert"
-          :post-handlers '(sp-latex-insert-spaces-inside-pair))
-
-  ;; some common wrappings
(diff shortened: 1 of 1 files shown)


Fuco1:
Hmm, to be fair though, the "default" behaviour is fine until you use those special language quotes, right?  I wonder if we shouldn't keep it working e.g. in english.

dschrempf:
For me, `smartparens` is completely broken in LaTeX, also in English. I am not using German in my documents. It could be a Doom Emacs issue. I removed all the relevant changes (see the patch), but who knows.

PR #1152, opened 2023-03-27
Add python-ts-mode (treesit) as an alias for python-mode

zyxir:
This has been done for the `elixir-ts-mode` before.  Maybe every tree-sitter powered modes, like `bash-ts-mode` or `c-ts-mode`, should be added as well?


Fuco1:
Thanks!  I'm still on some ancient Emacs version so I don't have the treesitter support.  I don't know if there's some way to do this automatically for each mode, but people are contributing the fixes.  It's just me taking so long :( Sorry for that.

PR #1162, opened 2023-05-11
Support `ruby-ts-mode`

danielma:
As of Emacs 29, `ruby-mode` and `ruby-ts-mode` are derived from `ruby-base-mode`. Have smartparens define behavior from the base mode to work with both

### DIFF
--- a/smartparens-ruby.el
+++ b/smartparens-ruby.el
@@ -293,9 +293,9 @@ MS, ME, MB."
         (sp-ruby-method-p "end"))))
 
 (add-to-list 'sp-navigate-skip-match
-             '((ruby-base-mode enh-ruby-mode motion-mode) . sp--ruby-skip-match))
+             '((ruby-mode ruby-ts-mode enh-ruby-mode motion-mode) . sp--ruby-skip-match))
 
-(dolist (mode '(ruby-base-mode motion-mode))
+(dolist (mode '(ruby-mode ruby-ts-mode motion-mode))
   (add-to-list 'sp-sexp-suffix `(,mode syntax "")))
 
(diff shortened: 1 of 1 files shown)


Fuco1:
This is a bit surprising but inheritance actually does not work everywhere in smartparens.  I don't remember if this was a conscious decision (probably wasn't) and I think it should be fixed.  But for now, your patch would not work.

Where it does work is pair configuration.

smartparens-ruby.el
```diff
@@ -293,12 +293,12 @@ MS, ME, MB."
         (sp-ruby-method-p "end"))))
 
 (add-to-list 'sp-navigate-skip-match
-             '((ruby-mode enh-ruby-mode motion-mode) . sp--ruby-skip-match))
+             '((ruby-base-mode enh-ruby-mode motion-mode) . sp--ruby-skip-match))
```
Fuco1, line 296: Here inheritance does not work so you need to mention each mode separately.

smartparens-ruby.el
```diff
@@ -293,12 +293,12 @@ MS, ME, MB."
...
 
-(dolist (mode '(ruby-mode motion-mode))
+(dolist (mode '(ruby-base-mode motion-mode))
```
Fuco1, line 298: Here inheritance does not work so you need to mention each mode separately.

Fuco1 (changes requested):

danielma:
@Fuco1 Thank you for the feedback! That does surprise me, and it seemed to work correctly for me locally, but I've fixed it now.

It was very helpful of you to point out the specific places I needed to change

Fuco1 (approved):

Fuco1:
Sorry for taking so long to merge this :blush:

PR #1167, opened 2023-07-11
Add commands named `sp-(backward-)?-delete-sexp`

ROCKTAKEY:
Hi. Thanks for maintenance.
 I love this project. I cannot write lisp without smartparens!

While functions which delete symbol and word exist like`sp-delete-symbol` and `sp-delete-word`, There is no function delete sexp. So I add it by this PR.

Thanks!


Fuco1:
Yea, this is reasonable, thanks for catching the missing versions!  Sorry for being so late with merging this :blush:

Issue #1115, opened 2021-11-28
sp-backward-delete-word causes an error `args-out-of-range` when the kill ring is empty


cordelia04m:
## Expected behavior
There should be no error when calling `sp-backward-delete-word` when the kill ring is empty.

## Actual behavior
An error happened and it took me into recursive edit.

## Steps to reproduce the problem
1. Clean the kill ring using this function
```
(defun clean-kill-ring ()
  "Use this to clean password after pasting it from a clipboard."
  (interactive)
  (setq kill-ring nil)
  (garbage-collect))
```
2. Call `sp-backward-delete-word`.
3. Call `sp-backward-delete-word` again. The error will happen this second time we call it.

## Backtraces if necessary (`M-x toggle-debug-on-error`)
```
Debugger entered--Lisp error: (args-out-of-range 0 0)
  get-text-property(0 yank-handler nil)
  kill-append(#("aaa " 0 3 (fontified t)) t)
  kill-region(6 3)
  sp-backward-kill-symbol(1 t)
  sp-backward-delete-symbol(1 t)
  sp-backward-delete-word(1)
  funcall-interactively(sp-backward-delete-word 1)
  call-interactively(sp-backward-delete-word nil nil)
  command-execute(sp-backward-delete-word)
```

## Environment & version information

- `smartparens` version: 20211101.1101
- Active `major-mode`: `lisp-interaction-mode`
- Smartparens strict mode: nil
- Emacs version (`M-x emacs-version`): GNU Emacs 27.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.27, cairo version 1.17.4) of 2021-03-27
- Starterkit/Distribution: Vanilla
- OS: gnu/linux

woolsweater:
I believe I know why this happens. I've had a related issue with deleting unexpectedly adding to the kill-ring. The `sp-delete-*` commands try to avoid touching the `kill-ring` by let-binding it before using kill commands. On repeated deletes/kills the `kill-region` call chain eventually ends up in a call to `add-to-history`: `(add-to-history 'kill-ring string kill-ring-max t)`. `add-to-history` looks up `kill-ring`'s value with `symbol-value`, which *bypasses lexical binding*. So it ends up unexpectedly mangling the actual global `kill-ring` despite the let binding.

[...]

Issue #1097, opened 2021-08-13
Delete commands bugging out yank


AkashaP:
Ive noticed a lot of delete commands seem to bug out the emacs yanking system, but im not sure exactly why. I dont know if this is because of smartparens but using smartparens based commands does cause the bug to happen.

## Expected behavior
delete commands to leave next yank alone, as per the steps:
1. write some text
2. Place cursor before it
3. invoke vanilla 'kill-region' command on it
4. Repeat 1 and 2 with different text
5. invoke vanilla 'delete-region' command on that
6. yank. Should yank the text that was killed using kill-region

## Actual behavior
The equivalent smartparens commands seem to bug out yank system completely and yank something i think i appended to the kill ring like days or weeks ago

## Steps to reproduce the problem

(The bug seems to happen with all delete-like commands and not just sp-delete-word)

1. write some text
2. Place cursor before it
3. invoke 'sp-kill-word' command on it
4. Repeat 1 and 2 with different text
5. invoke vanilla 'sp-delete-word' command on that
6. yank

I usually end up with text i dont even know where it is coming from. I'm relatively new to emacs, I looked at the smartparens source and saw that it does something along the lines of shadowing the kill-ring in a let so that the real kill-ring doesnt actually get changed.

In addition i can't yank-pop after step 6 i get the error 'previous command was not a yank'

But when i evaluate (car kill-ring-yank-pointer) i can see that the deleted text is still there ~~, but not in (car kill-ring). I dont really understand why kill-ring-yank-pointer contains different things from kill-ring or why it bugs out yank. I might experiment around with my own yank command that just pops kill-ring-yank-pointer or something.
EDIT just looked at the documentation for emacs and i think most of this is actually working as intended but~~
doing the steps for the vanilla emacs commands doesnt append the deleted text to kill-ring-yank-pointer

By the way, looking at emacs source i can see that emacs itself is coded to put its delete commands in terms of kill commands but i find that logic so backwards, shouldnt it be the other way around?

## Environment & version information

im using spacemacs-base for my emacs configuration

- `smartparens` version: 20210529.1129
- Active `major-mode`: `emacs-lisp-mode`
- Smartparens strict mode: nil
- Emacs version (`M-x emacs-version`): GNU Emacs 26.3 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.14) of 2020-03-26, modified by Debian
- Starterkit/Distribution: Spacemacs
- OS: gnu/linux

[...]

Issue #1040, opened 2020-07-03
Do not overwrite clipboard contents on C-Backspace


phoe:
## Expected behavior

I want `C-Backspace` to **not** overwrite the most recent thing I have pasted in the clipboard.

## Actual behavior

Hitting `C-Backspace` puts the deleted characters into the kill ring, overwriting whatever I had in the clipboard.

## Steps to reproduce the problem

Hit `C-Backspace` and then try to paste.

## Environment & version information
- `smartparens` version: 20200324.2147
- Active `major-mode`: `slime-repl-mode`
- Smartparens strict mode: nil
- Emacs version (`M-x emacs-version`): GNU Emacs 26.3 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.20) of 2020-05-17, modified by Debian
- Starterkit/Distribution: Spacemacs
- OS: gnu/linux

phoe:
Fixed by remapping `C-Backspace` to `sp-backward-delete-word`.

phoe:
Actually *not* fixed. Issuing `(define-key smartparens-mode-map (kbd "<C-backspace>") 'sp-backward-delete-word)` in IELM causes erratic behavior, as the kill ring is seemingly cycled over.

Fuco1:
Fixed in #1169

PR #1169, opened 2023-07-29
`kill-ring` should not be `let-bind`ed to `kill-ring`itself on `sp(-backward)?-delete-symbol`

ROCKTAKEY:
Hi, I fixed the bug that `sp-delete-symbol` and `sp-backward-delete-symbol` destruct `kill-ring`.
Here is reproduction process:
```emacs-lisp
(setq kill-ring '("kill-ring-value"))
; => ("kill-ring-value")
abc def ; At the beginning of this line, run `sp-delete-symbol' twice
kill-ring
; => (#("kill-ring-valuedef" 15 18 (fontified t)))
```

# Explanation
Wrapping with `(let ((kill-ring kill-ring) (select-enable-clipboard nil)) ...)` is no mean, because the `kill-*` commands destructively edit the list which globally binds `kill-ring`, even with `let-bind`.

`kill-ring` is just a list. When list is passed to other variable, it is NOT copied but passed as reference. So the `let-bind`ing `kill-ring` to `kill-ring` does nothing, and destructive editing  is not reverted. If we want to copy the list, you need to use `purecopy`.

Therefore, we can fix this bug by `let-bind`ing `kill-ring` to other list, like `nil`.

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -9085,7 +9085,7 @@ backward direction.
 
 See `sp-forward-symbol' for what constitutes a symbol."
   (interactive "*p")
-  (let* ((kill-ring kill-ring)
+  (let* ((kill-ring nil)
          (select-enable-clipboard nil))
     (sp-kill-symbol arg word)))
 
@@ -9162,7 +9162,7 @@ forward direction.
 
 See `sp-backward-symbol' for what constitutes a symbol."
   (interactive "*p")
-  (let* ((kill-ring kill-ring)
+  (let* ((kill-ring nil)
          (select-enable-clipboard nil))
     (sp-backward-kill-symbol arg word)))
 


woolsweater:
This is great! This will fix https://github.com/Fuco1/smartparens/issues/1115 and also, I believe https://github.com/Fuco1/smartparens/issues/1040 and https://github.com/Fuco1/smartparens/issues/1097 I've just tested it out on Emacs 29 and it seems to be working well there.

Fuco1:
This used to work because those operations did not use destructive edits :/  But yes, nil is definitely safer.  I don't know why I didn't go with that in the first place.

Fuco1:
Actually, this seems to be even more complicated, because there now is a variable `kill-ring-yank-pointer` which "points" to the kill ring, but if we replace the kill ring with nil, it will save a pointer to that locally bound list (since it's just a cons).  So we now need to also nil-bind this variable to shadow its dynamic binding.  I'm making a patch.

ROCKTAKEY:
Thanks, great work!!!

PR #1170, opened 2023-08-15
fix: don't indent after kill in python-ts-mode

bard:
Emacs 29 introduced treesitter programming modes. This updates the list of modes where a reindent shouldn't be performed after kill to take that into account.


Fuco1:
Thanks!

Issue #1168, opened 2023-07-22
Can't wrap JSX in `{` `}`


tien:
## Expected behavior

```jsx
// Be able to select the inner div and wrap it in  {}
<div>
  <div />
</div>

// Which will become
<div>
  {<div />}
</div>

// This is useful when doing something like this
<div>
  {somePredicate && <div />}
</div>
```

## Actual behavior

Receive the following error

```
mismatch sexp state: wrapping would break structure
```

## Environment & version information

- `smartparens` version: <Please specify manually>
- Active `major-mode`: `typescript-tsx-mode`
- Smartparens strict mode: nil
- Emacs version (`M-x emacs-version`): GNU Emacs 28.2 (build 1, aarch64-apple-darwin21.1.0, NS appkit-2113.00 Version 12.0.1 (Build 21A559)) of 2023-02-23
- Starterkit/Distribution: Doom
- OS: darwin

Fuco1:
Where does the `typescript-tsx-mode` come from?  I'm using the regular `typescript-mode` and it wraps.

PR #1174, opened 2023-10-21
Add support for js-ts-mode and typescript-ts-mode

akreisher:
Adds the new Emacs 29+ `js-ts-mode` and `typescript/tsx-ts-mode` to the configs for `smartparent-javascript`. 

I also added `tsx-ts-mode` to `sp--html-modes` which  should address https://github.com/Fuco1/smartparens/issues/1168.

### DIFF
--- a/smartparens-config.el
+++ b/smartparens-config.el
@@ -148,7 +148,7 @@ ID, ACTION, CONTEXT."
 (eval-after-load 'tuareg                   '(require 'smartparens-ml))
 (eval-after-load 'fsharp-mode              '(require 'smartparens-ml))
 (eval-after-load 'unisonlang-mode          '(require 'smartparens-unison))
-(--each '(js js2-mode)
+(--each '(js js2-mode typescript-ts-mode)
   (eval-after-load it                      '(require 'smartparens-javascript)))
 (provide 'smartparens-config)
 
--- a/smartparens-javascript.el
+++ b/smartparens-javascript.el
@@ -48,7 +48,16 @@
 
 ;; (|sys).path.append---the dot should not travel with the closing
 ;; paren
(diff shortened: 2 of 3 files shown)


smartparens-config.el
```diff
@@ -148,7 +148,7 @@ ID, ACTION, CONTEXT."
 (eval-after-load 'tuareg                   '(require 'smartparens-ml))
 (eval-after-load 'fsharp-mode              '(require 'smartparens-ml))
 (eval-after-load 'unisonlang-mode          '(require 'smartparens-unison))
-(--each '(js js2-mode)
+(--each '(js js2-mode typescript-ts-mode)
```
Fuco1, line 151: we should probably add `js-ts-mode` here as well then? Since you added it to the config.

Fuco1 (changes requested):

smartparens-config.el
akreisher: `js-ts-mode` is provided in `js.el`, so I believe just including `js` here should be sufficient for both `js-mode` and `js-ts-mode` (since I believe this form expects package names for `use-after-load`, not mode symbols). `typescript-ts-mode` on the other hand is in its own package (`typescript-ts-mode.el`) and so needs its own entry here.

I believe the same applies for python. This line:
https://github.com/fuco1/smartparens/blob/96fccc0816093aaad4b904a20802ec306c6da420/smartparens-config.el#L135-L136

could probably just be:

```
(eval-after-load 'python                   '(require 'smartparens-python))
```
since both `python-mode` and `python-ts-mode` are defined in `python.el`.

Fuco1: right, good catch!

Fuco1:
Thanks!

PR #1178, opened 2023-12-13
fix(clojure): remove single-quote pair, remove missing function ref

zilti:


zilti:
...ah isn't that great, now it doesn't insert `"` in pairs anymore. I think I'll just switch back to paredit.

PR #1180, opened 2023-12-13
feat: add rules for Emacs Lisp

zilti:


Fuco1:
I believe this is already configured in `smartparens-config.el`.

PR #1185, opened 2024-02-04
Add support for clojure-ts-mode

kommen:
[`clojure-ts-mode`](https://github.com/clojure-emacs/clojure-ts-mode)  is developed as a distinct alternative to `clojure-mode`, as discussed here: https://github.com/clojure-emacs/clojure-mode/issues/640#issuecomment-1359708131

### DIFF
--- a/smartparens-clojure.el
+++ b/smartparens-clojure.el
@@ -44,8 +44,7 @@
 (defvar sp-clojure-prefix "\\(?:[@`'#~,_?^]+\\)"
   "Prefix used in `sp-sexp-prefix' for clojure modes.")
 
-(dolist (mode '(clojure-mode clojurescript-mode clojurec-mode cider-repl-mode
-                clojure-ts-mode clojurescript-ts-mode clojurec-ts-mode))
+(dolist (mode '(clojure-mode cider-repl-mode clojure-ts-mode))
   (add-to-list 'sp-sexp-prefix `(,mode regexp ,sp-clojure-prefix)))
 
 ;; Match "`" with "`" in strings and comments
--- a/smartparens-clojure.el
+++ b/smartparens-clojure.el
@@ -44,7 +44,8 @@
 (defvar sp-clojure-prefix "\\(?:[@`'#~,_?^]+\\)"
   "Prefix used in `sp-sexp-prefix' for clojure modes.")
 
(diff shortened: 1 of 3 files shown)


kommen:
Marking as draft until this is resolved: https://github.com/clojure-emacs/clojure-ts-mode/pull/36

bbatsov:
Hmm, do we even need to iterate the derived modes if there's no special smartparens configuration for them? Seems excessive to me. Perhaps the legacy config is from the days before all modes were folded in `clojure-mode`?

kommen:
@bbatsov thanks for the feedback. I can give it a try in the next days without all the derived modes.

kommen:
I removed all the derived modes. As far as I can tell, everything works fine in buffers using the derived modes.

smartparens.el
```diff
@@ -562,9 +562,7 @@ Symbol is defined as a chunk of text recognized by
 (defcustom sp-lisp-modes '(
                            cider-repl-mode
                            clojure-mode
-                           clojurec-mode
-                           clojurescript-mode
-                           clojurex-mode
```
Fuco1, line 567: This needs to stay there because `sp-list-modes` is checked literally not taking into account how modes derive from one another.

The question is whether it should, but right now it doesn't, so this will actually remove these modes from being marked as "lisp" modes.

smartparens-clojure.el
```diff
@@ -44,7 +44,7 @@
 (defvar sp-clojure-prefix "\\(?:[@`'#~,_?^]+\\)"
   "Prefix used in `sp-sexp-prefix' for clojure modes.")
 
-(dolist (mode '(clojure-mode clojurescript-mode clojurec-mode cider-repl-mode))
```
Fuco1, line 47: This `sp-sexp-prefix` setting takes the modes literally, not taking into account how modes derive from one another.  So removing them here will break things unfortunatelly.

Fuco1 (changes requested):

kommen:
@Fuco1 I have reverted the commit which removed the modes from the lists. Would you prefer me to force push it out?

Fuco1:
I rebased these commits to master and merged them separately.  Sorry it took me so long.  I will close this pull request but don't worry as the changes made it in.

Thanks!

kommen:
@Fuco1 Thank you so much for your work and again for your feedback!

Issue #1201, opened 2024-04-10
Either fix ert failures or mark them as expected


temyurchenko:
## Expected behavior

Running `cask exec ert-runner` results in no failing tests.

## Actual behavior

First, `ert-runner` complains about duplicated test names. However, even after we fix those, we get 42 failures. The best and simplest way to deal with those is to mark them as [«expected failures»](https://www.gnu.org/software/emacs/manual/html_mono/ert.html#Expected-Failures). It is slightly complicated due to some tests being wrapped in macros.

## Steps to reproduce the problem

Run `cask build` followed by `cask exec ert-runner`. 

## Environment & version information

- `smartparens` version: 20231024.1804
- Active `major-mode`: `fundamental-mode`
- Smartparens strict mode: nil
- Emacs version (`M-x emacs-version`): GNU Emacs 29.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.40, cairo version 1.18.0)
- Starterkit/Distribution: Vanilla
- OS: gnu/linux

Fuco1:
The failing test cases need to be corrected, they should not be failing.  I am slowly fixing some but it's a long process.

temyurchenko:
> The failing test cases need to be corrected, they should not be failing. I am slowly fixing some but it's a long process.

Oh, yes, I absolutely get that. Sorry, if that sounded pressuring. I was just reading the contributing guide, which advised to run a test-suite, which I ran and it showed a lot of errors. I haven't found any documentation about that in the repo, so I decided to make an issue for others who might be confused.

Fuco1:
No worries :)

Fuco1:
I'll make a new issue with all the failing tests so I can track progress individually.

PR #1202, opened 2024-04-10
fix some of the tests; partially addresses #1201

temyurchenko:


temyurchenko:
> What is the new file that you created here for latex?

The new file is for testing `LaTeX-mode`, as opposed to `latex-mode`. There is a lot of difference in how they handle electricity, so I felt it's worth it to test both of the modes.

The SaRcAsM-case names belong to the AUCTeX package, while the normal-case names belong to the builtin tex-mode package.

In particular, the seems to be a confusion about this in the `sp--special-self-insert-commands` variable. The comment says:

    ;; At some point the TeX and LaTeX functions were renamed to
    ;; lower-case names.  This broke some code dealing with these
    ;; modes, so we just add both versions for now.

But they've never been renamed, they just belong to different packages. E.g. `TeX-insert-quote` is from AUCTeX, and `tex-insert-quote`. On the other hand, `TeX-insert-dollar` doesn't have a counterpart in the builtin tex-mode.el, so including it in the list seems excessive.

Fuco1:
Oh my, I never realized it's *that* stupid :D In that case yea it definitely makes sense to have tests for both modes, in fact ideally somehow "macroified" so we always test for both of them.

temyurchenko:
I have considered the idea of «macroifiying» them. I have decided to go against that in this PR for the following reasons

1. The modes are rather different. `LaTeX-mode` has a very different behaviour depending on the «electric» and other variables set and we might want to test against different combinations of those. They would be specific to each mode.
2. Macros kind of tie you to a syntax that is at times hard to extend. For example, I initially wanted to mark all of the current test failures as «expected failures» (https://www.gnu.org/software/emacs/manual/html_mono/ert.html#Expected-Failures) [1]. However, due to heavy macro usage in `sp-test sp-comment`, we would have to extend macro syntax in some way.

So, given that they are kind of different, and macros are compromise, and that there are «only» two duplicates, I felt that the current solution is not the worst point of the design space.


[1]: Why? Then, there would be an issue tracking the progress of fixing the failures. That way, we would still be able to run the test-suite regularly and track the regressions, whereas currently we might not notice that there is a new regression amidst those 42 failing tests.

Fuco1:
> [1]: Why? Then, there would be an issue tracking the progress of fixing the failures. That way, we would still be able to run the test-suite regularly and track the regressions, whereas currently we might not notice that there is a new regression amidst those 42 failing tests.

I didn't know it worked like that.  I'm not really happy with how the tests are written, over many years we used many different approaches, some good and some bad, so yea, it's a huge mess now.  It's one of my priorities now to get the test suite to pass everything, so I would rather focus on that since as you say the macro usage is quite heavy and makes things complicated.

Fuco1 (approved):

test/smartparens-LaTeX-test.el
jidicula, line 1: Sorry about the necropost ping, but this filename leads to some interesting `git` (and anything in the dev loop) behaviour on macOS, whose default APFS filesystem is case-insensitive by default (diffs between this file and `test/smartparens-latex-test.el` get merged and unmerged). Could this filename be changed to something different from the case-insensitive collision with `test/smartparens-latex-test.el`?

jidicula: I've opened a PR to address this https://github.com/Fuco1/smartparens/pull/1210

temyurchenko: Perhaps, it would be clearer to disambiguate by prefixing with «builtin-» and «auctex-» where appropriate?

jidicula: yeah that makes sense, I can modify my PR

Fuco1: Good catch and thanks for the PR, I merged it now :)

PR #1206, opened 2024-04-16
Fix invalid alias definition for `syntax-class-to-char'

### LINKED ISSUE #1204: Symbol definition void sp--syntax-class-to-char
## Expected behavior

Should be able to slurp code, e.g. slurp a collection into an expression

Before
```clojure
(map inc |) [1 2 3]
```
after
```clojure
(map inc [1 2 3]) 
```

> Slurp works with `smartparens-20231024.1804`

## Actual behavior

Code is not slurped and error in message buffer

```
sp-get-buffer-char-syntax: Symbol’s function definition is void: sp--syntax-class-to-char
```

Update to `smartparens-20240410.1709` package from melpa

Tested in latest Spacemacs develop branch and all latest packages downloaded as of 11 April 2024

Using `smartparens-20231024.1804` does not experience this issue (last version used).

**In recent enough `smartparens` you can call `M-x sp-describe-system` to generate this report**.  Please fill manually what we could not detect automatically.  Edit the output as you see fit to protect your privacy.

- `smartparens` version: 20240410.1709
- Active `major-mode`: `clojure-mode`
- Smartparens strict mode: nil
- Emacs version (`M-x emacs-version`): GNU Emacs 29.2 (build 1, aarch64-apple-darwin23.3.0, NS appkit-2487.40 Version 14.3 (Build 23D56)) of 2024-02-02
- Starterkit/Distribution: Spacemacs
- OS: darwin

juergenhoetzel:
`eval-when-compile' defines the alias symbol only at compile time and not when the program is loaded and thus results in a void error at runtime.

Fixes #1204

### DIFF
--- a/smartparens.el
+++ b/smartparens.el
@@ -1858,12 +1858,13 @@ P is the point at which we run `syntax-ppss'"
             (list p (point-min) (point-max))
             (sp-state-last-syntax-ppss-result sp-state) (syntax-ppss p)))))
 
-(eval-when-compile
-  (defalias 'sp--syntax-class-to-char 'syntax-class-to-char)
-  (when (version< emacs-version "28.1")
+;; `syntax-class-to-char' is new in Emacs 28.1.
+(defalias 'sp--syntax-class-to-char
+  (if (fboundp 'syntax-class-to-char)
+      #'syntax-class-to-char
     ;; Ripped from Emacs 27.0 subr.el.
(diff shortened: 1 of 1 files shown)


Fuco1:
Thanks!  I think you are right, I used a similar construction but the thing defined (for compat) was a macro so it was expanded during compile anyway and not necessary in runtime.  That's why it worked then but not in this case, good catch.

PR #1208, opened 2024-05-18
add LaTeX tests with delete-selection on

temyurchenko:
The tests fail on the current version of auctex, but pass on the most recent one following
https://git.savannah.gnu.org/cgit/auctex.git/commit/?id=45ee2f.

I'm not sure what's the best way to approach this is. 
AUCTeX used to have improper delete-selection properties on its «smart» functions. Even when the «smart» behaviour was turned of, the delete-selection property was overriden, causing smartparens functions to fail (see the tests). 

The patch fixing AUCTeX (unfortunately, not the quotes) has been accepted and is already in master. With that patch and this patch to smartparens, the tests will pass successfully.


Fuco1:
This is a great start but we need to make sure somehow to respect the older versions of Emacs / Auctex.  Maybe there is some condition that we can test (for example value of the delete-selection properties?) and only remove the commands from the "special" lists then, not always.  Given the rather slow release cycles of things surrounding Emacs, I think we need to be a bit careful not to remove something too soon.

temyurchenko:
> This is a great start but we need to make sure somehow to respect the older versions of Emacs / Auctex. Maybe there is some condition that we can test (for example value of the delete-selection properties?) and only remove the commands from the "special" lists then, not always. Given the rather slow release cycles of things surrounding Emacs, I think we need to be a bit careful not to remove something too soon.

Hey, that's reasonable. I've conditioned inclusion of TeX-insert-dollar (which is the only regression) on the auctex version.

PR #1210, opened 2024-06-08
Disambiguate LaTeX test filenames for case-insensitive filesystems

jidicula:
Fixes an issue where `smartparents-latex-test.el` previously conflicted with `smartparens-LaTeX-test.el` case-insensitive filesystems. See https://github.com/Fuco1/smartparens/pull/1202#discussion_r1632082004 for details.


Fuco1:
Thanks, this would also be problem on windows.

PR #1211, opened 2024-07-02
Use fboundp for functions

knu:
I'm not sure how to use sp-ruby-forward-sexp and sp-ruby-backward-sexp properly, but because I happened to find a misuse I'm submitting this.


Fuco1 (approved):

Fuco1:
As for how to use them, those functions are used internally so you don't use them on your own.

With the `fboundp` you are absolutely right, good catch!

PR #1213, opened 2024-07-03
 `RET` create a newline inside empty pairs 

0kaguya:
Rust prefers block indent style ([style guide](https://doc.rust-lang.org/stable/style-guide/index.html#block-indent)) so inserting a newline should be the default behavior.


Esnos33:
@0kaguya thank you for your code

PR #1216, opened 2024-07-13
config for c-ts-mode and c++-ts-mode

wyuenho:
Require `smartparens-c` when `c-ts-mode` is loaded, and add `c-ts-mode` and `c++-ts-mode` to `sp-c-modes`.


Fuco1:
Thanks!

PR #1221, opened 2024-12-14
Copy dependencies into smartparens.el for MELPA

marienz:
MELPA recently stopped extracting dependencies from smartparens-pkg.el (see
https://github.com/melpa/package-build/commit/760cf53114a6f27e86bc87b44726ced34bcda0ea). In a fresh install (using `emacs --init-directory` pointed at a directory with just an init.el configuring MELPA, no other packages installed), `M-x package-install smartparens` does not install dash, which means smartparens does not work.

Try to fix it by copying the dependencies from smartparens-pkg.el to a Package-Requires header, which MELPA should still extract.

Drop the dependency on cl-lib while I'm there. This was needed in Emacs < 24, which smartparens no longer supports.

### DIFF
--- a/smartparens-pkg.el
+++ b/smartparens-pkg.el
@@ -1,3 +1,2 @@
 (define-package "smartparens" "1.11.0" "Automatic insertion, wrapping and paredit-like navigation with user defined pairs."
-  '((dash "2.13.0")
-    (cl-lib "0.3")))
+  '((dash "2.13.0")))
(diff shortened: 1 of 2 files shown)


marienz:
Melpa rolled back their change, but still recommends migrating metadata to smartparens.el (see https://github.com/melpa/melpa/issues/9285#issuecomment-2543607077). So this PR is now a cleanup rather than a fix.

Fuco1:
Thanks for the patch / fix.  I include the pkg file for every one of my packages, so I guess I'll have to go and update everything.

PR #1224, opened 2025-01-27
Add go-ts-mode to smartparens-go.el

gav451:
Add `go-ts-mode` to `smartparens-go.el`.


Fuco1:
Thank you for the contribution!

Issue #1225, opened 2025-02-05
\( should not pair with \) in swift-mode


samglt:
Swift uses "\(...)" for interpolation in strings like this: `"Some Var: \(somevar)"` but the default sp behavior inserts a "\)" which is invalid. Just disabling the default pair with `(sp-local-pair "\\(" nil :actions nil)` fixes this. I'm not aware of any time parens are special characters that could need to be escaped in Swift.

I'll submit a PR with a test later today.

PR #1226, opened 2025-02-06
Don't pair escaped parens in swift-mode

samglt:
Related to #1225 

This disables pairing escaped parens together, but still allows a "\(" to be paired and become "\()" which is commonly used for string interpolation.

### DIFF
--- a/smartparens-swift.el
+++ b/smartparens-swift.el
@@ -106,7 +106,8 @@
   (sp-local-pair "<" ">"
                  :when '(sp-swift-filter-angle-brackets)
                  :skip-match 'sp-swift-skip-match-angle-bracket)
-  (sp-local-pair "\"\"\"" "\"\"\""))
+  (sp-local-pair "\"\"\"" "\"\"\"")
+  (sp-local-pair "\\(" nil :actions nil))
 
 ;; Swift has no sexp suffices.  This fixes slurping
 ;; (|foo).bar -> (foo.bar)
--- a/test/smartparens-swift-test.el
+++ b/test/smartparens-swift-test.el
@@ -123,3 +123,10 @@ func bar(x: UInt64) -> Bool {
     (mark-whole-buffer)
     (call-interactively 'sp-kill-region)
     (should (equal (buffer-string) ""))))
+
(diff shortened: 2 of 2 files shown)


woolsweater:
Nice, thanks for fixing this; I've been meaning to get around to it!

I would suggest a bit fuller of a definition:

```elisp
(sp-local-pair "\\(" ")" :actions '(insert wrap) :when '(sp-in-string-p))
```

Including the correct closing delimiter, rather than `nil`, lets other Smartparens facilities like `sp-unwrap-sexp` work correctly. The action list allows wrapping with a region. I.e., while it's a bit of an odd edit, if the region is active and denoted by the `|` characters:

```
let s = "foo |bar|"
```

then typing a backslash followed by an open paren correctly wraps "bar" in an interpolation.

Finally, while you're right that there's no case currently in Swift for escaped parens outside of a string, I don't think it hurts to include the `:when` predicate.

samglt:
Good points. Trying that out though, unwrapping only removes the parens and not the initial backslash, so `"\(|foo)"` becomes `"\|foo"`. The normal "(" ")" pairing has to take precedence I think

woolsweater:
Hmm, interesting 🤔 Are you using https://github.com/swift-emacs/swift-mode? I forgot actually that that mode does not propertize the interpolation delimiters such that smartparens can work with them (for example `sp-mark-sexp`, `sp-forward-sexp` etc. do not work). I just get an error trying to unwrap when using that mode.

swift-mode

Using https://codeberg.org/woolsweater/swift-ts-mode (authored by me :) does unwrap correctly:

swift-ts-mode

I have not, however, tried the other `swift-ts-mode`.

samglt:
That must be it. I'm using the [the other one](https://github.com/rechsteiner/swift-ts-mode). I'll try yours out

woolsweater:
This will be a bit of a tangle for smartparens. 🙁

Fuco1:
I think this is still a good change to include, even if in swift mode it doesn't work correctly, it doesn't break other functionality.  Plus it's probably more broken without it anyway :)

Fuco1:
Thank you for your contribution!

PR #1227, opened 2025-02-08
Add config for 'swift-ts-mode'

woolsweater:
Ensure `smartparens-swift` is loaded after `swift-ts-mode` in addition to `swift-mode`, and include `swift-ts-mode` in the customizations.


Fuco1:
Thank you for this improvement.  The addition of `-ts-` modes is a never ending battle :D

PR #1228, opened 2025-02-23
Fix erlang config

JimMoen:


Fuco1:
Thanks for the improvements!

